Issue #76, opened 2020-08-03
Could not find cn.webank.defibus:defibus-client:1.0.0


qqeasonchen:

lrhkobe:
bug has been fixed

PR #77, opened 2020-08-03
fix "can not find defibus-client"

qqeasonchen:
#76


MajorHe1:
Looking forward to more commits to fix hidden problems

keranbingaa:
It will lower the difficulty  for new starters.

PR #79, opened 2020-08-06
fix "can not find defibus-client"

qqeasonchen:


MajorHe1:
Good job, a important problem has been solved

PR #82, opened 2020-08-07
improve build scripts and fix known err

wqliang:
1. improve build scripts of root build script and emesher build script;
2. fix reading config value error;


qqeasonchen:
Very helpful

PR #85, opened 2020-08-07
link readme to store and emesher instruction

wqliang:
link readme of quickstart part to instruction of each components


qqeasonchen:
good job

Issue #97, opened 2020-08-08
getPID method get current process id improve speed


chenyi19851209:
getPID method get  current process id waste time,because of call 
ManagementFactory.getRuntimeMXBean() every time. call only once ok.

public static long getPID() {
        String processName = java.lang.management.ManagementFactory.getRuntimeMXBean().getName();
        if (processName != null && processName.length() > 0) {
            try {
                return Long.parseLong(processName.split("@")[0]);
            } catch (Exception e) {
                return 0;
            }
        }
        return 0;
    }

PR #98, opened 2020-08-08
get current process id only once

chenyi19851209:
@ #97 

get current process id only once，improve get current process id speed.


qqeasonchen:
good fix

PR #112, opened 2020-08-13
fix security leak

iNanos:


iNanos:
#111 #112 #109

qqeasonchen:
good job

Issue #128, opened 2020-08-19
improve start and stop script,  solve the problem deploying multiple process in the same server


lrhkobe:
improve start and stop script,  solve the problem deploying multiple process in the same server

PR #130, opened 2020-08-19
improve start and stop script

lrhkobe:
fix the problem when deploying multiple process of eventmesh-emesher  in same server #128


iNanos (commented):
LGTM

PR #178, opened 2020-12-14
[ISSUE #176] Support Spi for extended implementation to accommodate a variety of MQ

xwm1992:
issue #176


lrhkobe:
Split rocketMQ implementation from EventMesh  as SPI plugin

Issue #184, opened 2020-12-18
Eventmesh-connector-rocketmq can not send and receive msg successfully.


lrhkobe:

nguyenminhtuanfit:
<img width="1424" alt="image" src="https://github.com/apache/eventmesh/assets/6532409/7d87f56d-5d65-43c4-a158-0ac123793c06">
how to fix this issue

PR #185, opened 2020-12-18
Eventmesh-connector-rocketmq can not send and receive msg successfully.

xwm1992:
#184


lrhkobe:
if possible，do not submit so much files change in the same pull request

PR #187, opened 2020-12-23
adjust eventmesh-runtime config

lrhkobe:
adjust eventmesh-runtime config #186 

- remove dependency eventmesh-common in eventmesh-connector-api;
- separate eventmesh-runtime config, add plugin config file;
- fix manual ack problem.


xwm1992:
LGTM

PR #191, opened 2021-01-06
Support pub/sub ,heartbeat manage and subscribe manage for http

xwm1992:
#190


xwm1992:
update .md files and supply http sub, it's ok

Issue #211, opened 2021-02-03
Rocketmq-connector throw nullpointer exception


iNanos:
When I use rocketmq-connector with tcp sub pattern, it throw Nullpointer Exception, please have a look.

MajorHe1:
> When I use rocketmq-connector with tcp sub pattern, it throw Nullpointer Exception, please have a look.

ok，I'll take a look and handle this.

PR #214, opened 2021-02-05
[ISSUE #211]Rocketmq-connector throw nullpointer exception

xwm1992:
fix missing instanceName for broadcastConsumer refering to NullPointer Exception


keranbingaa (approved):
it's ok, supply the instance name for broadcast consumer when it is inited

Issue #212, opened 2021-02-05
  For multiple listening instances, only the last started instance can receive messages


mattmao0913:
Question desc:
For multiple listening instances, only the last started instance can receive messages

xwm1992:
ok, I'll take a look for this problem

PR #215, opened 2021-02-07
[ISSUE #212]For multiple listening instances, only the last started instance can receive messages

xwm1992:
fix：this problem because of heartBeat processor client info managing error

### DIFF
--- a/eventmesh-runtime/src/main/java/com/webank/eventmesh/runtime/core/protocol/http/processor/HeartBeatProcessor.java
+++ b/eventmesh-runtime/src/main/java/com/webank/eventmesh/runtime/core/protocol/http/processor/HeartBeatProcessor.java
@@ -150,18 +150,7 @@ public class HeartBeatProcessor implements HttpRequestProcessor {
                     proxyHTTPServer.localClientInfoMapping.put(groupTopicClientMapping.getKey(), groupTopicClientMapping.getValue());
                 }else {
                     List<Client> tmpClientList = groupTopicClientMapping.getValue();
-                    for (Client tmpClient : tmpClientList){
-                        boolean isContains = false;
-                        for (Client localClient : localClientList){
-                            if (StringUtils.equals(localClient.url, tmpClient.url)){
-                                isContains = true;
-                                break;
-                            }
-                        }
-                        if (!isContains){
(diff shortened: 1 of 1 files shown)


eventmesh-runtime/src/main/java/com/webank/eventmesh/runtime/core/protocol/http/processor/HeartBeatProcessor.java
```diff
@@ -145,7 +145,26 @@ public void processRequest(ChannelHandlerContext ctx, AsyncContext<HttpCommand>
         }
         synchronized (proxyHTTPServer.localClientInfoMapping){
             for (Map.Entry<String, List<Client>> groupTopicClientMapping : tmp.entrySet()) {
-                proxyHTTPServer.localClientInfoMapping.put(groupTopicClientMapping.getKey(), groupTopicClientMapping.getValue());
+                List<Client> localClientList =  proxyHTTPServer.localClientInfoMapping.get(groupTopicClientMapping.getKey());
+                if (CollectionUtils.isEmpty(localClientList)){
+                    proxyHTTPServer.localClientInfoMapping.put(groupTopicClientMapping.getKey(), groupTopicClientMapping.getValue());
+                }else {
+                    List<Client> tmpClientList = groupTopicClientMapping.getValue();
+                    for (Client tmpClient : tmpClientList){
```
iNanos, line 153:
```suggestion
                    for (Client tmpClient : tmpClientList){
                        boolean isContains = false;
                        for (Client localClient : localClientList){
                            if (StringUtils.equals(localClient.url, tmpClient.url)){
                                isContains = true;
                                break;
                            }
                        }
                        if (!isContains){
                            localClientList.add(tmpClient);
                        }
                    }
```
These code can be simplified and extracted as a method will be better

iNanos (approved):

eventmesh-runtime/src/main/java/com/webank/eventmesh/runtime/core/protocol/http/processor/HeartBeatProcessor.java
xwm1992: ok, I will fix that.

PR #269, opened 2021-04-13
Specification of code structure and file naming

wqliang:
1.rename proxy field to eventmesh
2.rename wemq and access to eventmesh


xwm1992 (approved):
LGTM

MajorHe1 (approved):
It  seems commits too many codes, which should be checked carefully with running.

PR #315, opened 2021-04-25
Create .asf.yaml

qqeasonchen:


.asf.yaml
```diff
@@ -0,0 +1,44 @@
...
+    - event-driven
+    - event-streaming
+    - event-sourcing
+    - event-governance
+    - cloud-native
+    - serverless
+    - workflow-serverless
+    - esb
+    - message-bus
+    - CQRS
```
kezhenxu94, line 33: I remember capital characters are not allowed in the labels.

.asf.yaml
```diff
@@ -0,0 +1,44 @@
...
+  enabled_merge_buttons:
+    squash:  true
+    merge:   false
+    rebase:  false
+  protected_branches:
+    master:
```
kezhenxu94, line 39: Do you want to switch the protected branch from develop to master?

keranbingaa (approved):

.asf.yaml
qqeasonchen: I think it is ok.

.asf.yaml
qqeasonchen: It is.

xwm1992 (approved):

.asf.yaml
kezhenxu94: > I think it is ok.

By “not allowed”, I mean it causes the settings failed to take effect

.asf.yaml
qqeasonchen: > Do you want to switch the protected branch from develop to master?

If i cancel the protection from all branches, committers can modify the code directly or only PMC? i did not get any instructions from official docs. thanks.

kezhenxu94: > > Do you want to switch the protected branch from develop to master?
> 
> If i cancel the protection from all branches, committers can modify the code directly or only PMC? i did not get any instructions from official docs. thanks.

Committers and (P)PMC members. People with write access to the repo can directly push to it.

qqeasonchen: > > > Do you want to switch the protected branch from develop to master?
> > 
> > 
> > If i cancel the protection from all branches, committers can modify the code directly or only PMC? i did not get any instructions from official docs. thanks.
> 
> Committers and (P)PMC members. People with write access to the repo can directly push to it.

Got，thanks.

PR #319, opened 2021-04-27
[ISSUE #322] Rename package name "com.webank.eventmesh" to "org.apache.eventmesh"

SteveYurongSu:
Fix issue #322 .

### DIFF
--- a/eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/configuration/CommonConfiguration.java
+++ b/eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/configuration/CommonConfiguration.java
@@ -15,10 +15,10 @@
 // * limitations under the License.
 // */
 //
-//package org.apache.runtime.configuration;
+//package com.webank.runtime.configuration;
 //
 //import com.google.common.base.Preconditions;
-//import org.apache.runtime.util.EventMeshUtil;
+//import com.webank.runtime.util.EventMeshUtil;
 //import org.apache.commons.lang3.StringUtils;
 //
 //public class CommonConfiguration {
--- a/eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/configuration/ConfigurationWraper.java
+++ b/eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/configuration/ConfigurationWraper.java
@@ -15,7 +15,7 @@
 // * limitations under the License.
 // */
 //
-//package org.apache.runtime.configuration;
+//package com.webank.runtime.configuration;
 //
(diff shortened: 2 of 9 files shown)


eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/configuration/CommonConfiguration.java
```diff
@@ -15,10 +15,10 @@
 // * limitations under the License.
 // */
 //
-//package org.apache.runtime.configuration;
+//package com.webank.runtime.configuration;
 //
 //import com.google.common.base.Preconditions;
-//import org.apache.runtime.util.EventMeshUtil;
+//import com.webank.runtime.util.EventMeshUtil;
```
xwm1992, line 21: here still com.webank

eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/configuration/ConfigurationWraper.java
```diff
@@ -15,7 +15,7 @@
 // * limitations under the License.
 // */
 //
-//package org.apache.runtime.configuration;
+//package com.webank.runtime.configuration;
```
xwm1992, line 18: Similar to the above

eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/domain/ConsumeRequest.java
```diff
@@ -14,7 +14,7 @@
 // * See the License for the specific language governing permissions and
 // * limitations under the License.
 // */
-//package org.apache.runtime.domain;
+//package com.webank.runtime.domain;
```
xwm1992, line 17: Similar to the above

xwm1992 (changes requested):

eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/configuration/CommonConfiguration.java
SteveYurongSu: Yes, I have noticed that. But I think there is no "org.apache.runtime" package...

SteveYurongSu: If the class is useless, we'd better remove it :D

xwm1992: Ok, you can remove these classes later

xwm1992 (approved):

Issue #337, opened 2021-05-10
Http.demo Subscriber is not working in EventMesh-test project


jinrongluo:
Http.demo Http Subscriber (SubService) is not working.  When the Async Http Producer (AsyncPublishInstance)  publishes a message to EventMesh runtime, This Subscriber is not able to consume it.

Steps to reproduce this issue in JAVA IDE Intellij:

1.  Import eventmesh project to Intellij and build the project.

2. Start up EventMesh runtime from eventmesh-starter project's  StartUp class

3. Start up Http Subscriber Test Application, from eventmesh-test project: org.apache.eventmesh.http.demo.sub.SpringBootDemoApplication.java

4. Start up Http Async Producer Test Application, from eventmesh-test project: 
org.apache.eventmesh.http.demo.AsyncPublishInstance.java

5. Observe the logs of Http Subscriber, it is not getting the published message from Eventmesh, Expecting log is 
  "=======receive message======= {message}"

6. I saw this error message in the EventMesh runtime. Which indicates the subscriber endpoint (http://10.48.244.96:8088/sub/test) is not available.

```
[...]

PR #338, opened 2021-05-10
[Issue #337] Fix Http Test Subscriber startup issue by moving the Thread.sleep into the child thread

jinrongluo:
Fix issue #337 

Http Test Subscriber is not able to startup due to the Thread.sleep is blocking the main startup thread.  Moving the Thread.sleep code inside the internal child thread (stopThread) will fix this issue.


vongosling:
This pr may solve this problem. But I would like to address some points: 1. Try your best to keep away the sleep calling in any case. In this context, you'd better make produce and consume in one case, when countdown latch is zero, you could call System.exit. 2. As for Runtime calling, you could replace it with context.registerShutdownHook() to delegate spring bean lifecycle management.

jinrongluo:
> This pr may solve this problem. But I would like to address some points: 1. Try your best to keep away the sleep calling in any case. In this context, you'd better make produce and consume in one case, when countdown latch is zero, you could call System.exit. 2. As for Runtime calling, you could replace it with context.registerShutdownHook() to delegate spring bean lifecycle management.

Thanks @vongosling Using countdown latch is a good suggestion for avoiding Thread.sleep.

PR #343, opened 2021-05-11
[Issue #337] Enhance Http Demo Subscriber by using ExecutorService, CountDownLatch and PreDestroy hook

jinrongluo:
Enhance Http Demo Subscriber by using ExecutorService, CountDownLatch and PreDestroy hook

- Remove Thread.sleep()
- Leveraging ExecutorService to manage the Thread lifecycle.
- Leveraging PreDestroy annotation to manage the JVM Runtime Shutdown hook

### DIFF
--- a/eventmesh-test/src/main/java/org/apache/eventmesh/http/demo/sub/controller/SubController.java
+++ b/eventmesh-test/src/main/java/org/apache/eventmesh/http/demo/sub/controller/SubController.java
@@ -41,11 +41,11 @@ public class SubController {
     @RequestMapping(value = "/test", method = RequestMethod.POST)
     public String subTest(@RequestBody String message) {
         logger.info("=======receive message======= {}", JSONObject.toJSONString(message));
+        subService.consumeMessage(message);
+
         JSONObject result = new JSONObject();
         result.put("retCode", 1);
-        String strResult = result.toJSONString();
-        subService.consumeMessage(strResult);
-        return strResult;
+        return result.toJSONString();
     }
 
 }
--- a/eventmesh-test/src/main/java/org/apache/eventmesh/http/demo/sub/controller/SubController.java
+++ b/eventmesh-test/src/main/java/org/apache/eventmesh/http/demo/sub/controller/SubController.java
@@ -19,8 +19,10 @@ package org.apache.eventmesh.http.demo.sub.controller;
 
 import com.alibaba.fastjson.JSONObject;
 
+import org.apache.eventmesh.http.demo.sub.service.SubService;
 import org.slf4j.Logger;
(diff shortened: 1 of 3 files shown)


eventmesh-test/src/main/java/org/apache/eventmesh/http/demo/sub/service/SubService.java
```diff
@@ -38,6 +44,11 @@
     final String dcn = "FT0";
     final String subsys = "1234";
 
+    // CountDownLatch size is the same as messageSize in AsyncPublishInstance.java (Publisher)
+    private CountDownLatch countDownLatch = new CountDownLatch(AsyncPublishInstance.messageSize);
+
+    private ExecutorService executorService = Executors.newFixedThreadPool(5);
```
xwm1992, line 50: Of cource it won't make a difference, but I wonder here is it necessary to use a thread pool?

eventmesh-test/src/main/java/org/apache/eventmesh/http/demo/sub/controller/SubController.java
```diff
@@ -33,12 +35,17 @@
...
+    private SubService subService;
+
     @RequestMapping(value = "/test", method = RequestMethod.POST)
     public String subTest(@RequestBody String message) {
         logger.info("=======receive message======= {}", JSONObject.toJSONString(message));
         JSONObject result = new JSONObject();
         result.put("retCode", 1);
-        return result.toJSONString();
+        String strResult = result.toJSONString();
+        subService.consumeMessage(strResult);
```
xwm1992, line 47: The semantics of this line of code looks like it should pass in the received message as a parameter, not the result of consumption

eventmesh-test/src/main/java/org/apache/eventmesh/http/demo/sub/service/SubService.java
jinrongluo: It is a best practice to use the JDK Concurrency library (ExecutorService + ThreadPool) to manage the lifecycle of Thread.

Just suggestion. I can rollback to create Thread manually. Thanks for review comment.

eventmesh-test/src/main/java/org/apache/eventmesh/http/demo/sub/controller/SubController.java
jinrongluo: Good comment! I will make the change to pass the message instead of result.

jinrongluo:
hi @xwm1992 please review my latest commits and let me know if any questions. Thanks.

xwm1992 (approved):
LGTM

Issue #344, opened 2021-05-11
Racing Condition Observed after multiple subscribe / unsubscribe calls


jinrongluo:
I am testing the  Http LiteConsumer of EventMesh using Eventmesh-Test project. 

**Problem Description**

I modify the ` org.apache.eventmesh.http.demo.sub.service.SubService.java ` to perform a serial of Subscribe / unsubscribe operations, as the following:

```
        liteConsumer = new LiteConsumer(eventMeshClientConfig);
        liteConsumer.start();
        liteConsumer.heartBeat(topicList, url);

        for(int i = 0; i < 10; i++) {
            liteConsumer.subscribe(topicList, url);
            Thread.sleep(3000);
            liteConsumer.unsubscribe(topicList, url);
            Thread.sleep(3000);
        }
```

I start the org.apache.eventmesh.http.demo.sub.SpringBootDemoApplication.java for the testing. During the SubService run, I observe the following exception in the EventMesh runtime log:

```
2021-05-11 13:51:00,848 DEBUG [eventMesh-http-worker-1] http(AbrstractHTTPServer.java:287) - httpCommand={REQ,POST/HTTP,requestCode=203,opaque=73,header=heartbeatRequestHeader={code=203,language=JAVA,version=V1,env=P,region=,idc=FT,dcn=FT0,sys=1234,pid=0,ip=127.0.0.1:60141,username=userName,passwd=password},body=heartbeatRequestBody={clientType=SUB}}
2021-05-11 13:51:00,855 DEBUG [eventMesh-http-worker-1] http(AbrstractHTTPServer.java:287) - httpCommand={REQ,POST/HTTP,requestCode=206,opaque=74,header=subscribeRequestHeader={code=206,language=JAVA,version=V1,env=P,region=,idc=FT,dcn=FT0,sys=1234,pid=0,ip=127.0.0.1:60141,username=userName,passwd=password},body=subscribeBody{url='http://10.48.245.117:8088/sub/test', topics=[FT0-e-80010001-01-1]}}
2021-05-11 13:51:01,078 INFO  [eventMesh-clientmanage-2] http(SubscribeProcessor.java:67) - cmd=SUBSCRIBE|http|client2eventMesh|from=127.0.0.1:60141|to=10.48.245.117
[...]

PR #345, opened 2021-05-11
[Issue #344] Fixing racing condition issue in SubscribeProcessor and UnSubscribeProcessor

jinrongluo:
- Fixing racing condition issue in SubscribeProcessor and UnSubscribeProcessor
- Fix Subscribe and Unsubscribe Client API


jinrongluo:
> Do not change it. checkstyle will be designed to disallow to use * .to import.

Sorry I was not intent to change the import statements.  Let me rollback that part of code. Thanks.

jinrongluo:
@vongosling please review my latest commits and let me know if you have any questions. Thanks.

eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/core/protocol/http/processor/UnSubscribeProcessor.java
```diff
@@ -131,12 +126,15 @@ public void onResponse(HttpCommand httpCommand) {
...
+
+            registerClient(unSubscribeRequestHeader, consumerGroup, unSubTopicList, unSubscribeUrl);
+
             for (String unSubTopic : unSubTopicList) {
                 List<Client> groupTopicClients = eventMeshHTTPServer.localClientInfoMapping.get(consumerGroup + "@" + unSubTopic);
                 Iterator<Client> clientIterator = groupTopicClients.iterator();
                 while (clientIterator.hasNext()) {
                     Client client = clientIterator.next();
-                    if (StringUtils.equals(client.ip, ip)) {
+                    if (StringUtils.equals(client.ip, ip) && StringUtils.equals(client.url, unSubscribeUrl)) {
```
xwm1992, line 137: **client.ip** here should change to use **client.pid** as for equals condition，here ip represent ip+port and the port is random this equal condition will never reached, and client won't unsubscribe

xwm1992 (changes requested):

eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/core/protocol/http/processor/UnSubscribeProcessor.java
jinrongluo: Thanks @xwm1992 I fixed it by using client.pid instead. Thanks.

jinrongluo:
@SteveYurongSu made a commit for CodeQL causing build break. Please investigate. Thanks.

SteveYurongSu:
> @SteveYurongSu made a commit for CodeQL causing build break. Please investigate. Thanks.

Thanks for reporting this.

I checked the log and saw that all Github Actions (including the greeting actions) had encounted internal errors, which was very abnormal.
image

So I checked the Github Status (https://www.githubstatus.com/history), and found that the Github Action service was down at that time. This may cause your runs failed.
image

The incident has been resolved. So please push a commit to trigger the checks and see the result. Let me know if the issue is still. 😊

xwm1992 (approved):
LGTM

Issue #368, opened 2021-05-26
Racing Condition problem in EventMesh SDK's LiteConsumer HTTP Client


jinrongluo:
EventMesh SDK's Http LiteConsumer is using HttpClient to make a RPC to EventMesh Runtime. But the HttpClient is static and shared by all LiteConsumer instances (as Class variable). There are two problems for using static variable here:

1. when multiple LiteConsumer instances are sharing the same HttClient, when one LiteConsumer shutdown(), the other LiteConsumer cannot make any RPC calls. That's because shutdown() will invoke httpClient.close();

2.  I also observed racing condition of using this HttpClient. That's because the HttpClient is not close() and clean up its resource until LiteConsumer.shutdown(). This HttpClient is managed by PooledHttpClientManager, and it is meant to be used by multi-thread process, and needs to close() after finishing using it.

To reproduce the racing condition issue. I slightly modify the `eventmesh-test` project's `org.apache.eventmesh.http.demo.sub.service.SubService` class, with the followings

```
        liteConsumer = new LiteConsumer(eventMeshClientConfig);
        liteConsumer.start();

        liteConsumer.subscribe(topicList, url);
        Thread.sleep(30000);
        liteConsumer.unsubscribe(topicList, url);
        Thread.sleep(30000);
        liteConsumer.subscribe(topicList, url);
        Thread.sleep(30000);
        liteConsumer.unsubscribe(topicList, url);
```

The relevant log is attach in a file.

[SubServiceLog_issue368.txt](https://github.com/apache/incubator-eventmesh/files/6548673/SubServiceLog_issue368.txt)

When examine the response time coming from the four RPC calls above. we can see the racing condition is happening.

1.  liteConsumer.subscribe(topicList, url);  The response time is matching with the Log time
[...]

PR #369, opened 2021-05-26
[Issue #368] Fix Racing condition and memory leak issue in EventMesh SDK LiteConsumer and LiteProducer

jinrongluo:
- Remove static variables in LiteConsumer and LiteProducer
 
- Enhance shutdown() method to properly close the object resources, to avoid memory leak

- Surround each RPC call with local HttpClient object and close it immediately after using it. Avoid memory leak an racing condition in multi-threading process


eventmesh-sdk-java/src/main/java/org/apache/eventmesh/client/http/consumer/LiteConsumer.java
```diff
@@ -126,10 +127,13 @@ public boolean subscribe(List<String> topicList, String url) throws Exception {
         long startTime = System.currentTimeMillis();
         String target = selectEventMesh();
         String subRes = "";
+        CloseableHttpClient httpClient = setHttpClient();
```
ruanwenjun, line 130: Can we use `try with resource` here?
```java
try (CloseableHttpClient httpClient = setHttpClient()){

}catch(){

}
```
And I think it's not very good to do a conversion on `Exception` here, which will hide the source of the `Exception`. If the `Exception` is not currently handled, the best way may to throw it directly instead of wrapping it.

eventmesh-sdk-java/src/main/java/org/apache/eventmesh/client/http/AbstractLiteClient.java
```diff
@@ -46,4 +53,22 @@ public LiteClientConfig getLiteClientConfig() {
     public void shutdown() throws Exception {
         logger.info("AbstractLiteClient shutdown");
     }
+
+    public CloseableHttpClient setHttpClient() throws Exception {
```
ruanwenjun, line 57: Declare that throwing a specific exception may be better than throw Exception.

jinrongluo: TrustManager[] tm = new TrustManager[]{new MyX509TrustManager()}; 

This line is throwing generic Exception. So it might not be appropriate to declare specific Exception in the method signature.

eventmesh-sdk-java/src/main/java/org/apache/eventmesh/client/http/consumer/LiteConsumer.java
jinrongluo: Thanks for your suggestion. I changed to use `try with resource`.

eventmesh-sdk-java/src/main/java/org/apache/eventmesh/client/http/AbstractLiteClient.java
```diff
@@ -46,4 +53,22 @@ public LiteClientConfig getLiteClientConfig() {
...
+        if (!liteClientConfig.isUseTls()) {
+            return HttpClients.createDefault();
+        }
+        SSLContext sslContext = null;
+        try {
+            String protocol = System.getProperty("ssl.client.protocol", "TLSv1.1");
```
yuzhoumao, line 63: is there any particular reason we need to default to TLS 1.1 here?

would it be better if we default to TLS 1.2 instead?

eventmesh-sdk-java/src/main/java/org/apache/eventmesh/client/http/AbstractLiteClient.java
yuzhoumao: since like you said the constructor `MyX509TrustManager()` throws a generic exception, we can wrap it into an `EventMeshException` object (`throw new EventMeshException(e);`)

this way we can just declare that `EventMeshException` would be thrown

eventmesh-sdk-java/src/main/java/org/apache/eventmesh/client/http/AbstractLiteClient.java
jinrongluo: This is the original code from the SDK. I believe TLS 1.1 is out dated. I can move it TLS 1.2 instead. Thanks for catching it.

eventmesh-sdk-java/src/main/java/org/apache/eventmesh/client/http/AbstractLiteClient.java
jinrongluo: @ruanwenjun has a comment to in other part of my PR, mentioning not to wrap generic exception into EventMeshException. I am fine with either way, actually.

yuzhoumao: I see

ruanwenjun: In fact, this is my opinion, feel free to discuss.

xwm1992 (approved):

Issue #428, opened 2021-07-09
The Chinese existing in the project was changed to English


Jackzeng1224:
Under these modules, there are some Chinese。To English?
![image](https://user-images.githubusercontent.com/35483870/125054152-7aa31d00-e0d8-11eb-88a1-73b7938e8a97.png)

Roc-00:
I can help translate it into English.

PR #431, opened 2021-07-10
[ISSUE #428]Fix Chinese annotation

Roc-00:
Translate the Chinese to English on the common module.
Only the Chinese part of the notes has been translated, and the Chinese in the code has not changed.


eventmesh-common/src/main/java/org/apache/eventmesh/common/protocol/http/body/client/HeartbeatResponseBody.java
```diff
@@ -27,13 +27,13 @@
 
 public class HeartbeatResponseBody extends Body {
 
-    //响应码
+    //react code
```
xwm1992, line 30: return code may be more appropriate

eventmesh-common/src/main/java/org/apache/eventmesh/common/protocol/http/body/message/ReplyMessageResponseBody.java
```diff
@@ -27,13 +27,13 @@
 
 public class ReplyMessageResponseBody extends Body {
 
-    //响应码
+    //react code
```
xwm1992, line 30: same as above

eventmesh-common/src/main/java/org/apache/eventmesh/common/protocol/http/body/message/ReplyMessageResponseBody.java
```diff
@@ -27,13 +27,13 @@
...
     private Integer retCode;
 
-    //响应信息
+    //react message
```
xwm1992, line 33: response message

eventmesh-common/src/main/java/org/apache/eventmesh/common/protocol/http/body/message/SendMessageBatchResponseBody.java
```diff
@@ -27,13 +27,13 @@
 
 public class SendMessageBatchResponseBody extends Body {
 
-    //响应码
+    //react code
```
xwm1992, line 30: same as above

eventmesh-common/src/main/java/org/apache/eventmesh/common/protocol/http/body/message/SendMessageBatchResponseBody.java
```diff
@@ -27,13 +27,13 @@
...
     private Integer retCode;
 
-    //响应信息
+    //react message
```
xwm1992, line 33: same as above

eventmesh-common/src/main/java/org/apache/eventmesh/common/protocol/http/body/message/SendMessageBatchResponseBody.java
xwm1992, line 30: same as above

eventmesh-common/src/main/java/org/apache/eventmesh/common/protocol/http/body/message/SendMessageBatchResponseBody.java
xwm1992, line 33: same as above

eventmesh-common/src/main/java/org/apache/eventmesh/common/protocol/http/body/message/SendMessageBatchV2ResponseBody.java
```diff
@@ -27,13 +27,13 @@
 
 public class SendMessageBatchV2ResponseBody extends Body {
 
-    //响应码
+    //react code
```
xwm1992, line 30: same as above

eventmesh-common/src/main/java/org/apache/eventmesh/common/protocol/http/body/message/SendMessageBatchV2ResponseBody.java
```diff
@@ -27,13 +27,13 @@
...
     private Integer retCode;
 
-    //响应信息
+    //react code
```
xwm1992, line 33: same as above

xwm1992 (changes requested):

eventmesh-common/src/main/java/org/apache/eventmesh/common/protocol/http/body/client/HeartbeatResponseBody.java
ruanwenjun: I think we may not need to add comments for each attribute, for example here, the `retCode` can express variable information. And I would probably prefer to rename it to `returnCode` rather than add a comment.

MajorHe1 (approved):

eventmesh-common/src/main/java/org/apache/eventmesh/common/protocol/http/common/ClientRetCode.java
```diff
@@ -21,6 +21,7 @@
 
     /**
      * 这个RETRY的意思是 客户端发现投递的消息它没有监听时, 告诉EventMesh 发往下一个, 重试几次以实现灰度 , 预留
```
ruanwenjun, line 23: We may also need to remove this Chinese comment

eventmesh-common/src/main/java/org/apache/eventmesh/common/protocol/http/common/ProtocolKey.java
```diff
@@ -24,7 +24,7 @@
     public static final String VERSION = "Version";
 
     public static class ClientInstanceKey {
-        ////////////////////////////////////协议层请求方描述///////////
+        ////////////////////////////////////Protocol layer requester description///////////
```
ruanwenjun, line 27: It seems not to be a good practice using `///////////////////////` in code.

eventmesh-common/src/main/java/org/apache/eventmesh/common/protocol/http/common/ClientRetCode.java
Roc-00: Got it.

eventmesh-common/src/main/java/org/apache/eventmesh/common/protocol/http/common/ProtocolKey.java
Roc-00: what about this“//Protocol layer requester description//”?

ruanwenjun: I think it might be better not use symbols in comments, the book 《Clean code》 suggests that we should use the simplest comments.
But this is just programming style.

xwm1992 (approved):

Issue #374, opened 2021-05-31
Improve code coverage


qqeasonchen:
It is only 6% currently which can be see here https://app.codecov.io/gh/apache/incubator-eventmesh, need to add unit test.

PR #440, opened 2021-07-15
[ISSUE #374]add test case.

tydhot:
Fixes ISSUE#<XXXX>.

### Motivation

*Explain the content here.*

### Modifications

### Documentation

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -123,7 +123,8 @@ subprojects {
     ]
 
     List junit = [
-            "junit:junit:4.12"
+            "junit:junit:4.12",
+            "com.github.stefanbirkner:system-rules:1.16.1"
     ]
 
     List log = [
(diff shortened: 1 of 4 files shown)


build.gradle
```diff
@@ -123,7 +123,8 @@ subprojects {
     ]
 
     List junit = [
-            "junit:junit:4.12"
+            "junit:junit:4.12",
+            "com.github.stefanbirkner:system-rules:1.16.1"
```
ruanwenjun, line 127: This dependency is use [`Common Public License 1.0 `](https://github.com/stefanbirkner/system-rules/blob/master/LICENSE), I think we can use here, but this need PPMC to confirm.
By the way, I don't see this jar used at your pr, if it is not used, I don't recommend adding this dependency.

tydhot: It is used to simulation docker ip in env variable .

tydhot: btw, I think it would be ok to use this library. https://github.com/search?q=org%3Aapache+com.github.stefanbirkner&type=code
It is widely used in other apache project now.

ruanwenjun: Cool

Issue #442, opened 2021-07-17
Fix findings filtered by Checkstyle workflow


SteveYurongSu:
### Is your enhancement request related to a problem? Please describe

I introduced CheckStyle workflow in PR #438, but did not fix the issues that Checkstyle found.

### Describe the solution you'd like

I will fix all findings filtered by the Checkstyle workflow.

Issue #438, opened 2021-07-15
Enable Github Actions for Checkstyle


SteveYurongSu:
### Is your enhancement request related to a problem? Please describe

I noticed that there are folks working on issue #428 (#431), translating or removing Chinese chars in code.
Manually checking for this kind of code style can be a pain.
But, we can make the process of code style check automated with less pain by introducing Checkstyle into CI.

### Describe the solution you'd like

I'd like to define a Checkstyle workflow in CI.
The Checkstyle workflow will fail if a bad code style is found and will provide a report after each run. 
The report includes the code style issues found in the workflow, and it can be a reference for reviewers.

### Describe alternatives you've considered

* Github Actions: Checkstyle check + reviewdog report 
* Github Actions: Spotless check

### Additional context

reviewdog: https://github.com/reviewdog/reviewdog
checkstyle-idea: https://plugins.jetbrains.com/plugin/1065-checkstyle-idea

SteveYurongSu:
Yes, we can also check our code style in IDE.

![image](https://user-images.githubusercontent.com/30497621/125798875-78b4649a-4a8e-4746-8172-337b42367fbd.png)

SteveYurongSu:
Hi @qqeasonchen @xwm1992 , my PR is ready, please take a look 😊😊
BTW, This PR may help #428 (#431).

PR #433, opened 2021-07-13
[ISSUE #367] Refactor connector

ruanwenjun:
1.support load plugin from eventMeshPluginDir
2.remove connector plugin from runtime

In local develop, we can add dependency of `eventmesh-connector-rocketmq` in `build.gradle`.
And remove the plugin when packing.

PR #443, opened 2021-07-17
[ISSUE #442] Fix findings filtered by Checkstyle workflow

SteveYurongSu:
Fixes ISSUE #442.

### Motivation

*Explain the content here.*

I introduced CheckStyle workflow in PR #438, but did not fix the issues that Checkstyle found.

### Modifications

* Fixed all findings filtered by the Checkstyle workflow.

* Modified Checkstyle config file:
```xml
<module name="RegexpSingleline">
    <property name="format" value="System\..+\.println"/>
    <property name="message" value="Prohibit invoking System.*.println in source code !"/>
</module>
```


ruanwenjun:
I also do the same thing at #433, if this pr is merged first, I will rebase.

qqeasonchen:
It seems this pr needs be merged first or every other new pr will be checked failed.

ruanwenjun:
@qqeasonchen Yes, If there are no other problems, this pr can be merged first, otherwise  the other commits will be blocked.

qqeasonchen:
Have any other guys reviewed already?

ruanwenjun (approved):
LGTM, When I submitted a pr, I found that the file `sChat2.jks` seems to break the rules in checkStyle, do we need to exclude this file? I am not sure whether this will happen.

SteveYurongSu:
> LGTM, When I submitted a pr, I found that the file `sChat2.jks` seems to break the rules in checkStyle, do we need to exclude this file? I am not sure whether this will happen.

@ruanwenjun Thanks for reviewing! 

I have the same question. It seems that `sChat2.jks` is an unexpected binary file, should we remove the file? @qqeasonchen

ruanwenjun:
> > LGTM, When I submitted a pr, I found that the file `sChat2.jks` seems to break the rules in checkStyle, do we need to exclude this file? I am not sure whether this will happen.
> 
> @ruanwenjun Thanks for reviewing!
> 
> I have the same question. It seems that `sChat2.jks` is an unexpected binary file, should we remove the file? @qqeasonchen

If this file is not needed, we can remove it, and if this is needed then we can filter it like below:
```java
<module name="BeforeExecutionExclusionFileFilter">
    <property name="fileNamePattern" value="./eventmesh-runtime/conf/sChat2.jks$"/>
  </module>
``` 
Or using [checkStyle-suppressions.xml](https://github.com/checkstyle/checkstyle/blob/master/config/suppressions.xml)

qqeasonchen:
sChat2.jks can be removed.

PR #450, opened 2021-07-21
[ISSUE #442] update checkStyle file

ruanwenjun:
Since the `sChat2.jks` has been removed, we can remove the filter.

### DIFF
--- a/style/checkStyle.xml
+++ b/style/checkStyle.xml
@@ -51,8 +51,4 @@
     </module>
   </module>
 
-  <module name="BeforeExecutionExclusionFileFilter">
-    <property name="fileNamePattern" value="./eventmesh-runtime/conf/sChat2.jks$"/>
-  </module>
-
 </module>


ruanwenjun:
@SteveYurongSu Hi, `sChat2.jks` has been added again, do you think we still need to remove this filter?

SteveYurongSu:
> @SteveYurongSu Hi, `sChat2.jks` has been added again, do you think we still need to remove this filter?

I have noticed that. @qqeasonchen What is the file used for? 😂

ruanwenjun:
@SteveYurongSu This file is used for ssl configure, it can be changed by user.
https://github.com/apache/incubator-eventmesh/blob/4eaebde2506a582f8c211143b8d22dedf21e4f5c/eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/boot/SSLContextFactory.java#L44-L55

SteveYurongSu:
> @SteveYurongSu This file is used for ssl configure, it can be changed by user.
> https://github.com/apache/incubator-eventmesh/blob/4eaebde2506a582f8c211143b8d22dedf21e4f5c/eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/boot/SSLContextFactory.java#L44-L55

Thanks! 

I don't know if we should keep the binary file in our repo, especially when we don't know what content is in. 
Maybe documentation is needed for this file?
  > What content is in?
  > Or how to generate it.

ruanwenjun:
@SteveYurongSu Yes, there already has a doc about enable ssl, https://github.com/apache/incubator-eventmesh/blob/develop/docs/cn/features/https.zh-CN.md.
Maybe we can enhance this doc. So do you agree to keep this file, if this file is needed, we may need to filter this file in checkStyle file.

SteveYurongSu:
> Maybe we can enhance this doc. So do you agree to keep this file, if this file is needed, we may need to filter this file in checkStyle file.

@ruanwenjun Yes, I agree. Let's keep this file in repo and add a filter in our checkStyle file. 😊

Issue #405, opened 2021-06-29
[Task List] Looking for Code Hero for Apache EventMesh


xwm1992:
Welcome to participate bello tasks:
- [ ] [Easy to fix]https://github.com/apache/incubator-eventmesh/issues/579
- [ ] [Feature]https://github.com/apache/incubator-eventmesh/issues/417
- [ ] [Feature]https://github.com/apache/incubator-eventmesh/issues/553
- [ ] [Feature]https://github.com/apache/incubator-eventmesh/issues/676
- [ ] [Feature]https://github.com/apache/incubator-eventmesh/issues/577

or you can find the issue from issue list that no one take it and you wish to contribute, thanks.

xwm1992:
- #431

Pil0tXia:
Superseded by https://github.com/apache/eventmesh/issues/3873

PR #474, opened 2021-07-27
[ISSUE #405]maybe there is no need to use atomicboolean here

tydhot:
in this class, complete only change false to true and never will be false again. And when you want to use get() in atomicBoolean, is just return a volatile boolean. If we use volatile here instead atomicBoolean, we will reduce each meaningless CAS check and there's no difference with before.


ruanwenjun (approved):
LTGM, it is better to modify the title of this pr to [ISSUE #405]

tydhot:
got it, thanks

xwm1992 (approved):

Issue #476, opened 2021-07-29
Biz Exceptions occured in EventMesh cause connection close of client


lrhkobe:
### Describe the bug
When exception occured in the method of `channelRead0`,  it will be thrown, and handled by the method 'exceptionCaught' in the class of `EventMeshTcpExceptionHandler` .
```
public class EventMeshTcpMessageDispatcher extends SimpleChannelInboundHandler<Package> {

    @Override
    protected void channelRead0(ChannelHandlerContext ctx, Package pkg) throws Exception { 
        try {
            ...

            dispatch(ctx, pkg, startTime, cmd);
        } catch (Exception e) {
            logger.error("exception occurred while pkg|cmd={}|pkg={}|errMsg={}", cmd, pkg, e);
            //throw new RuntimeException(e);
            throw e;
        }
    }
}
```

When EventMesh caught the exception, it will send goodbye instruction to the client, that means the biz exception caused the connection close of client. It seems unreasonable.
```
public class EventMeshTcpExceptionHandler extends ChannelDuplexHandler {

    @Override
    public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) throws Exception {
        Session session = eventMeshTCPServer.getClientSessionGroupMapping().getSession(ctx);
        UserAgent client = session == null ? null : session.getClient();
        logger.error("exceptionCaught, push goodbye to client|user={},errMsg={}", client, cause.fillInStackTrace());
        String errMsg;
        if (cause.toString().contains("value not one of declared Enum instance names")) {
            errMsg = "Unknown Command type";
        } else {
            errMsg = cause.toString();
        }

        if (session != null) {
            EventMeshTcp2Client.goodBye2Client(eventMeshTCPServer, session, errMsg, OPStatus.FAIL.getCode(), eventMeshTCPServer.getClientSessionGroupMapping());
        } else {
            EventMeshTcp2Client.goodBye2Client(ctx, errMsg, eventMeshTCPServer.getClientSessionGroupMapping(), eventMeshTCPServer.getEventMeshTcpMonitor());
        }
    }
}
```

PR #477, opened 2021-07-29
[ISSUE #476] Biz Exceptions occured in EventMesh cause connection close of client 

lrhkobe:
Fixes ISSUE #476 

### Motivation

When the biz exception occured in EventMesh,  such as, if  throwing the ThreadPoolRejectException in the code of 
 `eventMeshTCPServer.getTaskHandleExecutorService().submit(task)` , it will cause the connection close of client. This looks obviously unreasonable.

### Modifications

Catch the biz exception of EventMesh, and send the exception msg to client, not close the connection of client


eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/core/protocol/tcp/client/EventMeshTcpMessageDispatcher.java
```diff
@@ -77,8 +76,45 @@ protected void channelRead0(ChannelHandlerContext ctx, Package pkg) throws Excep
             dispatch(ctx, pkg, startTime, cmd);
         } catch (Exception e) {
             logger.error("exception occurred while pkg|cmd={}|pkg={}|errMsg={}", cmd, pkg, e);
```
ruanwenjun, line 78: The `errMsg={}` in the error log is redundant. It is better midify to 
```java
logger.error("exception occurred while pkg|cmd={}|pkg={}", cmd, pkg, e);
```

lrhkobe: Thank you, I got it, and I will fix it later.

Issue #473, opened 2021-07-27
LiteProducer async request code is sync


ruanwenjun:
## Bug Report
In `LiteProducer.java`, you can find the below method:
https://github.com/apache/incubator-eventmesh/blob/737c0f09efeba61a62653d33fd553491af7d0deb/eventmesh-sdk-java/src/main/java/org/apache/eventmesh/client/http/producer/LiteProducer.java#L207-L245

Does this method want to send async request ? It seems that it send sync reqeust.

We may need to change the request code to `MSG_SEND_ASYNC`

xwm1992:
This is not bug in my opinion，the request method with rrcallback is using for send sync message, you can refer the protocol.md.

ruanwenjun:
Thanks

PR #479, opened 2021-08-01
[ISSUE #473]fix configuration property get error

ruanwenjun:
fix the configuration check error

### DIFF
--- a/eventmesh-common/src/main/java/org/apache/eventmesh/common/config/ConfigurationWrapper.java
+++ b/eventmesh-common/src/main/java/org/apache/eventmesh/common/config/ConfigurationWrapper.java
@@ -80,7 +80,7 @@ public class ConfigurationWrapper {
         if (StringUtils.isEmpty(configValue)) {
             return defaultValue;
         }
-        Preconditions.checkState(StringUtils.isNumeric(configKey), String.format("%s error", configKey));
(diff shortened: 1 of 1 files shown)


eventmesh-common/src/main/java/org/apache/eventmesh/common/config/ConfigurationWrapper.java
```diff
@@ -80,7 +80,7 @@ public int getIntProp(String configKey, int defaultValue) {
         if (StringUtils.isEmpty(configValue)) {
             return defaultValue;
         }
-        Preconditions.checkState(StringUtils.isNumeric(configKey), String.format("%s error", configKey));
+        Preconditions.checkState(StringUtils.isNumeric(configValue), String.format("%s error", configKey));
```
xwm1992, line 83: here String.format is still configKey?

ruanwenjun: Yes, do we need to use configValue?
or use `String.format("key:%s value:%s error", configKey, configValue)`

xwm1992: I prefer `String.format("key:%s value:%s error", configKey, configValue)` this.

ruanwenjun: Done

Issue #484, opened 2021-08-05
Support access control about security for EventMesh


lrhkobe:
### Describe the solution you'd like

1. Add `eventmesh-security-plugin` module refers to `eventmesh-connector-plugin`,  `eventmesh-security-plugin`  contains acl api design module and concrete acl plugin implementation module;
2. I will provide a simple API interface design in `eventmesh-security-api` module and  support access control in the `eventmesh-runtime` by calling the API interface.
3. In order to avoid eventmesh running error , I provide a  plugin( `eventmesh-security-acl` ) of the default empty implementation of the `eventmesh-security-api`, as to this plugin implemention, you can consider your own requirements.If you are interested in this issue, welcome to participate and suggestions.

For access control, we can provide several ways as follows：

- check client ip, which can be used in blacklist or whitelist
- check username and password of client
- check whether the pub/sub relation of the subsystem is registered,such as, it can used to check whether the subsystem can send or receive msg from the specified topic.

###  API interface design in `eventmesh-security-api`

```
@EventMeshSPI(isSingleton = true)
public interface AclService {
    void init() throws AclException;

    void start() throws AclException;

    void shutdown() throws AclException;

    void doAclCheckInConnect(Properties aclProperties) throws AclException;

    void doAclCheckInHeartbeat(Properties aclProperties) throws AclException;

    void doAclCheckInSend(Properties aclProperties) throws AclException;

    void doAclCheckInReceive(Properties aclProperties) throws AclException;

}
```

PR #485, opened 2021-08-05
[ISSUE #484] Support access control for EventMesh

lrhkobe:
Fixes ISSUE #484 

### Motivation

Add access control about security in EventMesh,  enhance the features of eventmesh.

### Modifications

1. Add `eventmesh-security-plugin` module;
2. Add sub module `eventmesh-security-api` in the `eventmesh-security-plugin`,  this module provides simple access control api design;
3. Add sub module `eventmesh-security-api` in the `eventmesh-security-plugin`, this module only provides default empty implementation of plugin according to `eventmesh-security-api` so far;
4. Add acl check code in eventmesh-runtime when client send command of connecting, heartbeting, sending msg and subscribing .


qqeasonchen:
@lrhkobe good job, but if your impl is one choice, then the moudle should not named as eventmesh-acl-impl, because it is very common.

qqeasonchen:
LGTM

lrhkobe:
> @lrhkobe good job, but if your impl is one choice, then the moudle should not named as eventmesh-acl-impl, because it is very common.

OK, I will reactor the module name later.

eventmesh-runtime/conf/eventmesh.properties
```diff
@@ -56,4 +56,8 @@ eventMesh.server.registry.fetchRegistryAddrIntervalInMills=20000
 #eventMesh.server.defibus.client.comsumeTimeoutInMin=5
 
 #connector plugin
-eventMesh.connector.plugin.type=rocketmq
\ No newline at end of file
+eventMesh.connector.plugin.type=rocketmq
+
+#acl plugin
+eventMesh.server.acl.enabled=false
+eventMesh.acl.plugin.type=acl
```
xwm1992, line 63: eventMesh.server.security.enabled & eventMesh.security.plugin.type maybe better

eventmesh-common/src/main/java/org/apache/eventmesh/common/config/CommonConfiguration.java
```diff
@@ -29,23 +29,13 @@
     public String eventMeshName = "";
     public String sysID = "5477";
     public String eventMeshConnectorPluginType = "rocketmq";
+    public String eventMeshAclPluginType = "acl";
```
xwm1992, line 32: eventMeshAclPluginType -> eventMeshSecurityPluginType is better

eventmesh-common/src/main/java/org/apache/eventmesh/common/config/CommonConfiguration.java
```diff
@@ -29,23 +29,13 @@
...
-    public Integer ackWindow = 1000;
-    public Integer pubWindow = 100;
-    public long consumeTimeout = 0L;
-    public Integer pollNameServerInteval = 10 * 1000;
-    public Integer heartbeatBrokerInterval = 30 * 1000;
-    public Integer rebalanceInterval = 20 * 1000;
     public Integer eventMeshRegisterIntervalInMills = 10 * 1000;
     public Integer eventMeshFetchRegistryAddrInterval = 10 * 1000;
     public String eventMeshServerIp = null;
+    public boolean eventMeshServerAclEnable = false;
```
xwm1992, line 38: like above

xwm1992 (changes requested):

eventmesh-runtime/conf/eventmesh.properties
lrhkobe: OK， I will adjust it.

eventmesh-common/src/main/java/org/apache/eventmesh/common/config/CommonConfiguration.java
lrhkobe: OK， I will adjust it.

xwm1992 (approved):

Issue #339, opened 2021-05-11
[Feature] Integrate With OpenSchema


qqeasonchen:
option 1. reference openmessage
https://github.com/openmessaging/openschema/issues/1
https://github.com/openmessaging/openschema/blob/master/spec.md

yzhao244:
I am a user of eventmesh and I'm extremely interested in contributing to this project. I will go through the project related to EventMesh Schema Registry implementation which integrates with OpenSchema.
I will follow them and get back to the community for further discussions.

jinrongluo:
OpenSchema spec preview is released here:

https://github.com/openmessaging/openschema/blob/master/spec.md

qqeasonchen:
> I am a user of eventmesh and I'm extremely interested in contributing to this project. I will go through the project related to EventMesh Schema Registry implementation which integrates with OpenSchema.
> I will follow them and get back to the community for further discussions.

welcome and looking forward to your contributions.

yzhao244:
What is the question:

I am attempting a design which integrates with OpenSchema and is also easy to extend. 

What would you like to be added:
I suggest to add two more modules in overall eventmesh projects

1. eventmesh-store-api 
  This is a interface module which contains schemas registry persistency APIs such as the followings. 
```
  public interface EventSchemaService extends SchemaRegistry  {
	void createSchema(SchemaRequest schemaRequest);

    List<SchemaResponse> readAllSchemas();
    
    void updateSchema(SchemaRequest schemaRequest, String schemaId);   
	
    void deleteSchema(SchemaRequest schemaRequest, String schemaId);
  }
```	
  

2. eventmesh-store-h2
   This module contains the actual implementation of EventSchemaService which integrates with OpenSchema. I proposed to leverage using h2 database for persisting schema registry in eventmesh.
   However, this is also a pluggable module. Therefore, vendors can implement persistency using other techniques such as file system or any other data stores at their own will.

   
Why is this needed:

It ensures extendibility of Schema Registry in eventmesh since vendors may have requirements of using different techniques such as in-memory db, mysql db or any other data store for persisting data.
Furthermore, store layer can be extended with other management infomation for persistency such as subscriptions, topics. It is just this time we do for schema registry. :)

qqeasonchen:
@yzhao244 This store here better differ from event-store in connector, what do you think?

qqeasonchen:
@yzhao244 Would you like to share some new designs for us? thanks.

yzhao244:
[...]

Issue #435, opened 2021-07-13
Expose clientmanager/admin APIs as RESTful APIs


yzhao244:
### Is your enhancement request related to a problem? Please describe

Yes,  the clientmanager/admin APIs which are exposed from eventmesh-runtime module are not following the best practice of RESTful APIs . The problems could be seen by API urls such as "showClient", "rejectAllClient" ,"redirectClientByPath"

### Describe the solution you'd like

Add a new module such as eventmesh-rest which optimizes the existing admin/clientmanager APIs and expose them as RESTful APIs. 

### Describe alternatives you've considered
I will follow up and provide more details later. However, I would like to report this to the community and initiate a discussion here. :) . Please provide your feedbacks.

qqeasonchen:
good suggestions, just go ahead.

qqeasonchen:
#346 #349 are related to this.

qqeasonchen:
@yzhao244 I think you can submit this pr first.

qqeasonchen:
@yzhao244 Is there any updates here?

yzhao244:
@qqeasonchen Yes,  Basically, my idea is to move httpserver context creation, handlers along with its dependencies such as Utils classes to eventmesh-rest module.  Inside the module, a new utility class called "RequestMapping" maps RESTful RequestUrl to the corresponding handler method and extracts values from path variables. As my sample handler class code indicates, any request invokes REST API "/schemaregistry/subjects/{subject}/versions" eventually calls createSchemaHandler method. Additionally, subject value is extracted from the path variable {subject} and passed to handler method as well.   

![image](https://user-images.githubusercontent.com/15642771/129267814-3e01eac8-d925-40db-bc87-0e53cc4cf28e.png)

yzhao244:
HttpServer is still created and run inside eventmensh-runtime. Only the server.createContext part and handlers move to eventmesh-rest module.

qqeasonchen:
@yzhao244 thanks, you can see the new pr #498 , maybe you can give some updates as well, after all, you are the first contributor of this feature :), i hope you can be continue on this, thanks again.

JunjieChou:
Hi, @yzhao244. I would like to coorperate with you on Open Schema related coding part. Any updates here?

yzhao244:
[...]

PR #498, opened 2021-08-12
[ISSUE #339] Design the architecture and interfaces of OpenSchema integration

JunjieChou:
# Schema Integration in EventMesh
## ISSUE
#339
## Goals
- ensure a message could be understood by both producer and consumer
- validate a message is serialized correctly

## Comparison of Schema Integration in EventMesh and Other Projects
Project | Application
:---: | :---
EMQ | use "Schema Registry" and "Rule Matching" to transfer a message from one serialization format to another
Pulsar | use "Schema Registry" to validate a message
Kafka(Confluent) | satisfy the aforementioned two goals, inside which the second one is optional
EventMesh | satisfy the aforementioned two goals

## Archetecture
EventMeshSchemaRegistryArchitecture_3rd_edition

## LifeCycle of Schema in Messages
The highlevel lifecycle of schema in messages undergoes 9 steps as follows:
- step1: Producer registers a schema to OpenSchema Registry through OpenSchema service plugin.
- step2: Producer receives schema id from OpenSchema Registry.
- step3: Producer patch the schema id in front of messages and send them to EventMesh.
- step4: In a load-balanced way, EventMesh validate whether the format of messages is correct. Here, the load-balanced way means that the EventMesh, which is the entry port of messages, validates messages one skiping one and the rest messages are validated in the EventMesh which is the outer port.
- step5: EventMesh patch a status(true/false) for validation in front of schema id and router it to the EventMesh where the outer port exists.
- step6: EventMesh validates those non-validated messages.
- step7: EventMesh unpatch validation status and send [schema id + message] to the consumer.
- step8: Consumer unpatch the schema id and retreive the schema from OpenSchema Registry.
- step9: Consumer de-serialize messages according to schema.
## OpenSchema Restful
Method | URL | RequestBody | ResponseBody
:--- | :--- | :--- | :---
**Subject**|
POST | /subjects/{string:subject} | {subject:{JSON:subject}} | {subject:{JSON:subject}}
GET | /subjects | None | {subjects:{JSONArray:subjects}}
GET | /subjects/{string:subject} | None | {subject:{JSON:subject}}
GET | /schemas/ids/{string:id}/subjects | None | {subject:{JSON:subject}}
DELETE | /subjects/{string:subject} | None | [versions:{JSONArray:versions}]
**Schema**|
POST | /subjects/{string:subject}/versions | {schema:{JSON:schema}} | {id:{string:id}}
GET | /schemas/ids/{string:id} | None | {schema:{JSON:schema}} 
GET | /subjects/{string:subject}/versions/{int:version}/schema | None | {schema:{JSON:schema}}
GET | /subjects/{string:subject}/versions | None | [versions:{JSONArray:versions}]
DELETE | /subjects/{string:subject}/versions/{int:version} | None | version:{int:version}
**Compatibility**|
POST | /compatibility/subjects/{string:subject}/versions/{int:version} | {schema:{JSON:schema}} | {is_compatible:{boolean:is_compatible}}
PUT | /config/{string:subject} | {"compatibility":{string:compatibility}} | {"compatibility":{string:compatibility}}
GET | /config/{string:subject} | None | {"compatibility":{string:compatibility}}

## Java Interface Definition
We define three beans(```Subject.java```, ```Schema.java```, and ```Compatibility.java```), and two interfaces(```OpenSchemaService.java``` and ```OpenSchemaValidationService.java```).
```java
public class Subject {
    String tenant;
    String namespace;
    String subject;
    String app;
    String description;
    String status;
    Compatibility compatibility;
    String coordinate;
}

public class Schema {
    Subject subject;
    String name;
    int id;
    String comment;
    String serialization;
    String schemaType;
    String schemaDefinition;
    String validator;
    int version;
}

public enum Compatibility {
    BACKWARD,
    BACKWARD_TRANSITIVE,
    FORWARD,
    FORWARD_TRANSITIVE,
    FULL,
    FULL_TRANSITIVE,
    NONE
}

@EventMeshSPI(isSingleton = false)
public interface OpenSchemaService {

    // Subject
    Subject registerSubject(String subjectName, Subject subject);

    List<Subject> retrieveAllSubjects();

    Subject retrieveSubjectByName(String subjectName);

    Subject retrieveSubjectBySchemaID(int schemaID);

    int[] deleteSubjectByName(String subjectName);

    // Schema
    int registerSchema(String subjectName, Schema schema);

    Schema retrieveSchemaByID(int id);

    Schema retrieveSchemaByVersionInSubject(String subjectName, int version);

    List<Schema> retrieveAllSchemasBySubject(String subjectName);

    int deleteSchemaByVersionInSubject(String subjectName, int version);

    // Compatibility
    boolean isSchemaCompatibleWithVersionInSubject(String subjectName, int version, Schema schema);

    Compatibility alterCompatibilityBySubject(String subjectName, Compatibility compatibility);

    Compatibility retrieveCompatibilityBySubject(String subjectName);
}

@EventMeshSPI(isSingleton = true)
public interface OpenSchemaValidationService {
    Schema retrieveSchemaByID(int id);

    boolean validateMessageFormatBySchema(String message, Schema schema);
}

```


jinrongluo:
The event-rest module interface looks good. Thanks.

settings.gradle
```diff
@@ -25,4 +25,5 @@ include 'eventmesh-test'
 include 'eventmesh-spi'
 include 'eventmesh-connector-plugin:eventmesh-connector-api'
 include 'eventmesh-connector-plugin:eventmesh-connector-rocketmq'
+include 'eventmesh-rest'
```
ruanwenjun, line 28: Maybe it is better to rename this module to eventmesh-rest-api like other spi module if you hope to add another implementations in the future.

eventmesh-rest/src/main/java/org/apache/eventmesh/rest/OpenSchema/SchemaService/OpenSchemaService.java
```diff
@@ -0,0 +1,57 @@
...
+import org.apache.eventmesh.rest.OpenSchema.Bean.Schema;
+import org.apache.eventmesh.rest.OpenSchema.Bean.Subject;
+import org.apache.eventmesh.spi.EventMeshSPI;
+
+import java.util.List;
+
+@EventMeshSPI(isSingleton = true)
+public interface OpenSchemaService {
+
+    // Subject
```
ruanwenjun, line 29: It is better to use stander java annotation in spi interface.
https://www.oracle.com/technical-resources/articles/java/javadoc-tool.html.

eventmesh-rest/build.gradle
```diff
@@ -0,0 +1,26 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+dependencies {
+    implementation project(':eventmesh-spi')
+    testImplementation 'org.junit.jupiter:junit-jupiter-api:5.7.0'
```
ruanwenjun, line 20: Maybe you can just use `org.junit.jupiter:junit-jupiter-api` here, the version of dependency is defined in build.gradle of root project

ruanwenjun (commented):
BTW some URIs seem not meet the REST specification.
For example 
image

eventmesh-rest/build.gradle
JunjieChou: > Maybe you can just use `org.junit.jupiter:junit-jupiter-api` here, the version of dependency is defined in build.gradle of root project

OK

eventmesh-rest/src/main/java/org/apache/eventmesh/rest/OpenSchema/SchemaService/OpenSchemaService.java
JunjieChou: Hi, thanks for your review, what standard annotation do you think I should use?  I set this annotation the same as annotation of  ```org.apache.eventmesh.api.consumer.MeshMQPushConsumer``` because they are all plugins in runtime.

settings.gradle
JunjieChou: > Maybe it is better to split this module into two modules like `eventmesh-openschema-api` and `eventmesh-openschema-rest` like other spi module if you hope to add another implementations in the future.

That's a good suggestion. Should I create a new ```eventmesh-openschema-plugin``` module under main```EventMesh``` project, and put ```eventmesh-openschema-api``` and ```eventmesh-openschema-rest``` in it. Or create ```eventmesh-openschema-plugin``` module under ```eventmesh-rest```. I prefer the former one, but #435 suggests all RESTful APIs in ```eventmesh-rest```, how do you think?

ruanwenjun: I think the first one is good, since the module is main about openschema

eventmesh-rest/src/main/java/org/apache/eventmesh/rest/OpenSchema/SchemaService/OpenSchemaService.java
ruanwenjun: Yes, the annotation is right.
It's my wrong, I just mean you can add the standard comment annotation to write comment in interface, then others can  easily know the method of the interface, and add implementation easily.

JunjieChou (commented):
new commit

settings.gradle
yzhao244: I think @ruanwenjun 's idea of creating eventmesh-openschema-rest under its own plugin is good. Basically, the plugin becomes very self-contained.. Regarding #435, I will update to suggest creating a new module called eventmesh-admin-plugin which contains its own eventmesh-admin-rest along with other dependencies.

qqeasonchen:
@JunjieChou I sugget commiting the design doc seperate with code implementation. for  this pr, you can just commit the doc only.

PR #502, opened 2021-08-19
[ISSUE #339] add design doc for integrating OpenSchema

JunjieChou:


docs/en/features/eventmesh-schemaregistry-design.md
```diff
@@ -0,0 +1,71 @@
...
+## Design Details
+
+### Architecture
+
+![OpenSchema](https://user-images.githubusercontent.com/28994988/129255292-e61acc87-5250-4be5-ac9c-a099f6ef157c.png)
+
+### LifeCycle of Schema
+
+The highlevel lifecycle of schema in messages undergoes 9 steps as follows:
+- step1: Producer registers a schema to OpenSchema Registry through OpenSchema service plugin.
```
ruanwenjun, line 54: Can the producer register the schema by communicate with eventmesh-runtime admin server? In your description, the producer communicate with the schema plugin, I think the schema plugin just define the schema store and the schema type, they don't care how to communicate with client.

And the client should also only need to connect to eventmesh-runtime.

JunjieChou: As far as I was concerned, schema plugin provides (de-)serialization/validation/cache, and exists in both eventmesh-runtime and client sdk.

The reason for client connecting to Schema Registry directly is that it forms one-step to register a schema. Compared to client connecting to EventMesh and further to Schema Registry which is two-step, there is one step less. 
However, the former one requires EventMesh retrieving schema when producer-message arrives which could be omitted from the latter one due to cache during registering. So I think both are ok.

Of course the client could only connect to eventmesh-runtime if you think that's a necessity. And I will design this soon after.

qqeasonchen: I agree client interact with eventmesh-runtime only, it is more cloudnative

JunjieChou: Sure OK, I have committed new design.

ruanwenjun (approved):
LGTM, but there might one thing need to confirmed.

In this plan, the schema id will be added in each message. Does this will modify the current protocol? Maybe we should add the schema id in header or props?
Maybe the schema can be the metadata of a topic, then the producer doesn't need to send the schema id, and it also doesn't need to be notified when the scheme added/deleted in topics, since the validation and serialize is managed in eventmesh-runtime.

JunjieChou:
> LGTM, but there might one thing need to confirmed.
> 
> In this plan, the schema id will be added in each message. Does this will modify the current protocol? Maybe we should add the schema id in header or props?
> I think the schema can be the metadata of a topic, then the producer doesn't need to send the schema id, and it also doesn't need to be notified when the scheme added/deleted in topics, since the validation and serialize is managed in eventmesh-runtime.

As for the format of schema id, sure it could be added in the header as long as an MQ supports user-defined headers. Anyway, It's only about how to organize schema id which doesn't affect the architecture or processes.

As for where to generate schema id, It might work to put it on eventmesh-runtime, but I have one question:
How to set the format of messages(JSON?/JDK?/Avro?/...) between client and runtime? Maybe set it to JSON? because the current format is JSON and other types of format will introduce more problems😂.

qqeasonchen:
Mostly the design is ok, so i merge it frist, if there are still some detail discussion, please feel free to do it.

Issue #375, opened 2021-05-31
[Feature] In memory as event store


qqeasonchen:
For easy testing on eventmesh runtime only, there may need a in memorty event store.

vongosling:
+1, standalone mode.

TrumanDu:
+1

PR #507, opened 2021-08-22
[ISSUE #375] In memory as event store

ruanwenjun:
Add standalone connector plugin, close issue #375 

How to use:
1. Config the connector type in eventmesh.properties
```
#connector plugin
eventMesh.connector.plugin.type=standalone
```
2. install the standalone plugin
You can add the dependency in eventmesh-runtime/build.gradle
```
implementation project(":eventmesh-connector-plugin:eventmesh-connector-standalone")
```
Or execute 
```
./gradlew clean jar dist copyConnectorPlugin
```

### DIFF
--- a/eventmesh-connector-plugin/eventmesh-connector-standalone/src/main/java/org/apache/eventmesh/connector/standalone/broker/MessageQueue.java
+++ b/eventmesh-connector-plugin/eventmesh-connector-standalone/src/main/java/org/apache/eventmesh/connector/standalone/broker/MessageQueue.java
@@ -37,11 +37,11 @@ public class MessageQueue {
 
     private int count;
 
-    private ReentrantLock lock;
+    private final ReentrantLock lock;
 
-    private Condition notEmpty;
+    private final Condition notEmpty;
 
-    private Condition notFull;
+    private final Condition notFull;
 
 
     public MessageQueue() {
--- a/eventmesh-connector-plugin/eventmesh-connector-standalone/src/main/java/org/apache/eventmesh/connector/standalone/broker/MessageQueue.java
+++ b/eventmesh-connector-plugin/eventmesh-connector-standalone/src/main/java/org/apache/eventmesh/connector/standalone/broker/MessageQueue.java
@@ -1,0 +1,219 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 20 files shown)


eventmesh-connector-plugin/eventmesh-connector-standalone/src/main/java/org/apache/eventmesh/connector/standalone/broker/MessageQueue.java
```diff
@@ -0,0 +1,219 @@
...
+
+    public MessageEntity[] items;
+
+    private int takeIndex;
+
+    private int putIndex;
+
+    private int count;
+
+    private ReentrantLock lock;
```
tydhot, line 40: maybe the lock is final is better in this queue.

ruanwenjun: Thanks, agree with you.

xwm1992:
I try to use standalone mode to pub/sub. There is a `topic not exist exception` while client subscribe the topic, please have a look.

ruanwenjun:
@xwm1992 Thanks, I have fixed this problem, but it seems that the consumer has another problem, I am not sure.

xwm1992:
There is still have an exception during client subscribe to eventmesh runtime server, server didn't send the `subscribe response` cmd to the client, which cause the client operation time out.

ruanwenjun:
@xwm1992 OK, I have fixed.

Jackzeng1224:
@ruanwenjun   hello,When I test it will appear "SimpleSubClientImpl(SimpleSubClientImpl.java:169)-msg ignored|HEARTBEAT_RESPONSE|Package{header=Header{cmd=HEARTBEAT_RESPONSE, code=0, msg='success', seq='6462122308'), body=null}"

ruanwenjun:
@Jackzeng1224 Thanks, but this error also appears, when I am using rocketmq.

eventmesh-connector-plugin/eventmesh-connector-standalone/src/main/java/org/apache/eventmesh/connector/standalone/broker/task/SubScribeTask.java
```diff
@@ -0,0 +1,109 @@
...
+                                        // update offset
+                                        offset.incrementAndGet();
+                                        logger.info("message commit, topic: {}, current offset:{}", topicName, offset.get());
+                                        break;
+                                    case ReconsumeLater:
+                                        // don't update offset
+                                        break;
+                                    case ManualAck:
+                                        // update offset
+                                        offset.incrementAndGet();
```
xwm1992, line 80: Here is the ManualAck, shouldn't update the offset after commit the message, there is an acktask in runtime, only client call this task, the offset will be increment

xwm1992 (changes requested):

eventmesh-connector-plugin/eventmesh-connector-standalone/src/main/java/org/apache/eventmesh/connector/standalone/broker/task/SubScribeTask.java
ruanwenjun: OK, please take a look.

xwm1992 (approved):

Issue #505, opened 2021-08-20
Abandoned classes or methods are recommended to be replaced


Jackzeng1224:
## Question

Abandoned classes or methods are recommended to be replaced. Sometimes it will be abnormal

![image](https://user-images.githubusercontent.com/35483870/130196379-9dd5088f-de27-466f-9a15-3e2255278f31.png)

PR #510, opened 2021-08-24
[ISSUE #505]remove all RandomStringUtils.

tydhot:
Fixes ISSUE#<XXXX>.

### Motivation

*Explain the content here.*

### Modifications

### Documentation

### DIFF
--- a/eventmesh-sdk-java/src/test/java/org/apache/eventmesh/client/http/demo/AsyncPublishInstance.java
+++ b/eventmesh-sdk-java/src/test/java/org/apache/eventmesh/client/http/demo/AsyncPublishInstance.java
@@ -17,8 +17,8 @@
 
 package org.apache.eventmesh.client.http.demo;
 
-import org.apache.commons.lang3.RandomStringUtils;
 import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.text.RandomStringGenerator;
 import org.apache.eventmesh.client.http.conf.LiteClientConfig;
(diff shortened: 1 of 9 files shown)


eventmesh-sdk-java/src/test/java/org/apache/eventmesh/client/http/demo/AsyncPublishInstance.java
```diff
@@ -32,6 +32,9 @@
 
     public static Logger logger = LoggerFactory.getLogger(AsyncPublishInstance.class);
 
+    private final static RandomStringGenerator randomGenerator = new RandomStringGenerator.Builder()
```
ruanwenjun, line 35: It seems RandomStringGenerator is thread safe, so we can just hold one instance in our application.
Can we add a new utils or factory to Wrapper RandomStringGenerator.
```java
public class InnerStringUtils {

	/**
	 *  generate random number string
	 * @param length string length
	 * @return
	 */
	public static String generateRandomNumber(int length) {
		return RandomStringGeneratorHolder.instance.generate(length);
	}

	public static class RandomStringGeneratorHolder() {
		private static RandomStringGenerator instance = new RandomStringGenerator.Builder()
				.withinRange('0', '9').build();
	}

}
```

tydhot: yes, a util is better

ruanwenjun (approved):
LGTM

xwm1992 (approved):

PR #512, opened 2021-09-01
[ISSUE #454] implement ```/schemas``` controller

JunjieChou:
In this pr, we are going to implement a separated schema registry according to OpenSchema-spec. The techniques used are spring-boot and spring-data-jpa.

Currently, we have accomplished ```/schemas``` controller defined in openschema-spec. 

Welcome reviews and suggestions so that we could improve it!

#435 
#454 
#339


JunjieChou:
sorry, I did not notice that this pr is to eventmesh. It might be better to create to ```yuridev```, because we are coorperating on this work.

I am closing this pr and recreating it to ```yuridev```.

PR #517, opened 2021-09-10
Inefficient Usage of Java Collection

DittoTool:
Hi,

We find that there are several inefficient usages of Java Collections:

1. The contains method is invoked upon a list object in a loop. We recommend replacing it with a HashSet.
Random access can occur at several linkedlist objects, which run in linear time complexity. We recommend replacing them with Arraylist objects.
2. ArrayList is inserted before an iteration, while multiple memory reallocation might occur when the size of the list exceeds its capacity. We recommend replacing it with a LinkedList.
3. LinkedList is not inefficient when we perform random access upon it. We recommend replacing it with an ArrayList.

We discovered the above inefficient usage of containers by our tool Ditto. The patch is submitted. Could you please check and accept it? We have tested the patch on our PC. The patched program works well.

Bests

Ditto

### DIFF
--- a/eventmesh-common/src/main/java/org/apache/eventmesh/common/IPUtil.java
+++ b/eventmesh-common/src/main/java/org/apache/eventmesh/common/IPUtil.java
@@ -24,6 +24,7 @@ import java.net.SocketAddress;
 import java.net.SocketException;
 import java.net.UnknownHostException;
 import java.util.ArrayList;
+import java.util.LinkedList;
 import java.util.Enumeration;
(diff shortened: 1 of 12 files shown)


ruanwenjun (commented):
LGTM, but you need to change this pr push to develop branch.

DittoTool:
> LGTM, but you need to change this pr push to develop branch.

Thank you for your confirmation. I will do it.

DittoTool:
We have changed them from `master` to `develop` in #518. @ruanwenjun

Issue #523, opened 2021-09-16
Adding Flow control RateLimiter support in Http Message Send Processor


jinrongluo:
## Enhancement Request

In Eventmesh runtime, there is flow control using RateLimiter in TCP MessageTransfer processor, and HTTP Batch Message send processor. However flow control is missing in Http MessageSend (Sync & Async) procrssors.

I have implemented RateLimiter for Http Message Send processors, and would like to contribute.

I will create a PR later.

### Is your enhancement request related to a problem? Please describe

Adding RateLimiter to Http Message Send Processor.

Using property file to set the rate limit

### Describe the solution you'd like

Implementation is similar to Ratelimiter for TCP and HTTP Batch Send

jinrongluo:
@vongosling

qqeasonchen:
good question, you can refer to tcp ratelimiter implementation.

jinrongluo:
thanks @qqeasonchen I created PR - https://github.com/apache/incubator-eventmesh/pull/524

Please review. thanks.

jinrongluo:
PR merged, close issue.

PR #524, opened 2021-09-17
[Issue #523] Adding Flow control RateLimiter support in Http Message Send Processor

jinrongluo:
Issue #523 Adding Flow control RateLimiter support in Http Message Send Processor

### Motivation

In Eventmesh runtime, there is flow control using RateLimiter in TCP MessageTransfer processor, and HTTP Batch Message send processor. However flow control is missing in Http MessageSend (Sync & Async) procrssors.

I have implemented RateLimiter for Http Message Send processors, and would like to contribute.

### Modifications

Adding RateLimiter to Http Message Send Processor.

Using property file to set the rate limit


eventmesh-common/src/main/java/org/apache/eventmesh/common/protocol/http/common/EventMeshRetCode.java
```diff
@@ -39,7 +39,8 @@
     EVENTMESH_SUBSCRIBE_ERR(17, "eventMesh subscribe err"),
     EVENTMESH_UNSUBSCRIBE_ERR(18, "eventMesh unsubscribe err"),
     EVENTMESH_HEARTBEAT_ERR(19, "eventMesh heartbeat err"),
-    EVENTMESH_ACL_ERR(20, "eventMesh acl err");
+    EVENTMESH_ACL_ERR(20, "eventMesh acl err"),
+    EVENTMESH_HTTP_MES_SEND_OVER_LIMIT_ERR(20, "eventMesh http msg send over the limit, ");
```
xwm1992, line 43: this code 20 isn't right

xwm1992 (changes requested):

eventmesh-common/src/main/java/org/apache/eventmesh/common/protocol/http/common/EventMeshRetCode.java
jinrongluo: Thanks @xwm1992 I fix the error code to make it as 21.

Adding new commit to this PR

xwm1992 (approved):

PR #608, opened 2021-11-23
Fix standalone connector interface, fix example

ruanwenjun:


ruanwenjun:
@xwm1992 Please take a look.

PR #621, opened 2021-11-30
Change TCP Decoder and Encoder

ruanwenjun:
Since currently the TCP Encoder and Decoder is relied by runtime and sdk. So we don't move the Protocol plugin to Encoder and Decoder. In the current code, when sdk pub message to runtime, decoder will decoder the tcp package and deserialize the message to json string, and protocol plugin will deserialize the json string to target message.

And runtime publish message to sdk, the sdk also need to deserialize the message json string, this code will duplicate in protocol plugin and sdk, we can optimize this in the future.

We can not use JsonUtils to deserialize CloudEvent, since there are no NoArgsConstructor in CloudEvent, it might need to write code to do this.


ruanwenjun:
@xwm1992 Please take a look.

PR #640, opened 2021-12-09
Bugfix  ISSUE##639

jonyangx:
Fixes ISSUE##639

### Motivation

use specified type on new object.

### Modifications

Map<String, Map<String, Integer>> result = null;

if (!clientGroupMap.isEmpty()) {
            result = new HashMap<>(); 

### Documentation
No


ruanwenjun (approved):
LGTM.
It seems there are conflicts, but it does tell use which files have conflicts, it's strange.

vongosling:
The issue has been resolved:-)

PR #654, opened 2021-12-15
[MINOR] Upgrade log4j version to 2.16.0

ruanwenjun:
Since log4j2 2.16.0 has been released, and it solved some security issue, we need to upgrade log4j version before 1.3.0 release.
More detail can see https://logging.apache.org/log4j/2.x/security.html.


xwm1992 (approved):
LGTM

Issue #768, opened 2022-02-14
[Bug] HTTP Subscriber: Updating subscription information is not taking effect


jinrongluo:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/eventmesh/issues?q=is%3Aissue) and found no similar issues.

### Environment

Window

### EventMesh version

master

### What happened

In HTTP subscription, the changes in subscription will not trigger the `ConsumerGroupManager` to refresh. So the updates of the subscription is not taking effect.  

The `ConsumerGroupManager` still using the old subscription even it is already updated.

### How to reproduce

Run test using the http subscriber sample in `eventmesh-examples -> Http.deom.sub` package, with the following changes:

```
public void afterPropertiesSet() throws Exception {

  ....

 eventMeshHttpConsumer.subscribe(topicList, url);

 // after 10s, update the subscription with new topic
  Thread.sleep(10000);

  List<SubscriptionItem> topicList = Lists.newArrayList(
            new SubscriptionItem("TEST2-TOPIC-HTTP-ASYNC", SubscriptionMode.CLUSTERING, SubscriptionType.SYNC));
   eventMeshHttpConsumer.subscribe(topicList, url);

   ....
}
```

After running this test, the logs shows no consumer is started for topic `TEST2-TOPIC-HTTP-ASYNC`

Eventmesh Runtime Log

```

2022-02-14 15:00:19,434 DEBUG [eventMesh-http-worker-1] http(AbstractHTTPServer.java:381) - httpCommand={REQ,POST/HTTP,requestCode=206,opaque=1,header=subscribeRequestHeader={code=206,language=JAVA,version=V1,env=P,idc=FT,sys=1234,pid=18328,ip=127.0.0.1:51386,username=,passwd=},body=subscribeBody{consumerGroup='EventMeshTest-consumerGroup', url='http://192.168.2.31:8088/sub/test', topics=[SubscriptionItem{topic=TEST-TOPIC-HTTP-ASYNC, mode=CLUSTERING, type=ASYNC}]}}
2022-02-14 15:00:19,589 INFO  [pool-11-thread-1] ConsumerManager(ConsumerManager.java:78) - clientInfo check start.....
2022-02-14 15:00:19,608 INFO  [eventMesh-clientManage-] http(SubscribeProcessor.java:78) - cmd=SUBSCRIBE|http|client2eventMesh|from=127.0.0.1:51386|to=192.168.2.31
[...]

PR #769, opened 2022-02-14
[Issue #768] fix issue in update HTTP subscriber

jinrongluo:
<!--

Fixes ISSUE #768

### Motivation

This is to fix the bug of updating HTTP subscriber, and it is not taking effect.

### Modifications

update `ConsumerManager` notification method to use object.clone() instead of the same object reference.

### DIFF
--- a/eventmesh-common/src/main/java/org/apache/eventmesh/common/protocol/SubscriptionItem.java
+++ b/eventmesh-common/src/main/java/org/apache/eventmesh/common/protocol/SubscriptionItem.java
@@ -17,7 +17,11 @@
 
 package org.apache.eventmesh.common.protocol;
 
-public class SubscriptionItem implements Cloneable {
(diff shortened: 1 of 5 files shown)


eventmesh-common/src/main/java/org/apache/eventmesh/common/protocol/SubscriptionItem.java
```diff
@@ -66,6 +66,11 @@ public String toString() {
                 + ", type=" + type
                 + '}';
     }
+
+    @Override
+    public SubscriptionItem clone() {
```
zmt-Eason, line 71: I suggest implementing Serializable interface to implement deep clone instead of overriding Cloneable interface.

zmt-Eason (approved):

xwm1992 (approved):

Issue #658, opened 2021-12-16
[Enhancement] Support CloudEvents Webhook Specification


jinrongluo:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/eventmesh/issues?q=is%3Aissue) and found no similar issues.

### Enhancement Request

Eventmesh supports CloudEvents Specification. But there is a missing piece of work about Http Webhook for CloudEvents.

https://github.com/cloudevents/spec/blob/v1.0.1/http-webhook.md

To completely meet the CloudEvents Spec, I suggest implement Http Webhook for CloudEvents in EventMesh. 

### Describe the solution you'd like

Implement HTTP Webhook for CloudEvents Spec - https://github.com/cloudevents/spec/blob/v1.0.1/http-webhook.md

In HTTP Push -> AsyncHTTPPushRequest class

### Are you willing to submit PR?

- [x] Yes I am willing to submit a PR!

vongosling:
There are many constraints in this pr. IMO, the most important must-be item is WebHook-Request-Origin, we should care about this part.

jinrongluo:
Thanks I will contribute PR for this CloudEvents webhook spec

jinrongluo:
For implementing CloudEvents WebHook Spec, the following items required:

**Delivery Notification Request**

1. use HTTPS when invoking the webhook
2.  HTTP Header uses `Content-Type`
3. HTTP Header includes `WebHook-Request-Origin`

**Delivery Notification Response**
1. If delivery is accepted and processed, there is payload in response, Http return Code 200 OK / 201 Created, Response HTTP Header uses `Content-Type`
2. If delivery is accepted and processed and no payload, HTTP return code 201 Created / 204 No Content
3. If delivery is accepted and process status is unknown, HTTP return code 202 Accepted
4.  If the delivery is retired and the target returns 410 Gone, Eventmesh should not send further requests
5.  If the delivery target not able to process the request due to exceeding request rate limit, it should return 429 Too Many Requests. and include `Retry-After` header. Eventmesh should use this value for next retry
6. If the delivery is not accepted due to the notification format is not understood. it should return 415 Unsupported Media Type

PR #772, opened 2022-02-16
[Issue #658] Eventmesh Http Support CloudEvents Webhook spec

jinrongluo:
Fixes ISSUE #658

### Motivation

Eventmesh Http Support CloudEvents Webhook spec

### Modifications

Eventmesh Http Support CloudEvents Webhook spec, details of the requirements is here:

https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/http-webhook.md

https://github.com/apache/incubator-eventmesh/issues/658#issuecomment-1036576090


eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/core/protocol/http/push/HTTPClientPool.java
```diff
@@ -19,16 +19,40 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import com.google.common.collect.Lists;
 
 public class HTTPClientPool {
 
+    public Logger logger = LoggerFactory.getLogger(this.getClass());
+
     private List<CloseableHttpClient> clients = Lists.newArrayList();
```
ruanwenjun, line 56: It's better to use synchronizeList to avoid concurrent problem.

eventmesh-security-plugin/eventmesh-security-api/src/main/java/org/apache/eventmesh/api/auth/AuthService.java
```diff
@@ -0,0 +1,39 @@
...
+@EventMeshSPI(isSingleton = true, eventMeshExtensionType = EventMeshExtensionType.SECURITY)
+public interface AuthService {
+
+    void init() throws AuthException;
+
+    void start() throws AuthException;
+
+    void shutdown() throws AuthException;
+
+    Map getAuthParams() throws AuthException;
```
ruanwenjun, line 38: It seems the return result is always `Map<String, String>`?
It might be better to add generic type.
```suggestion
Map<String, String> getAuthParams() throws AuthException;
```

eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/util/WebhookUtil.java
```diff
@@ -0,0 +1,108 @@
...
+    private static final Map<String, AuthService> authServices = new ConcurrentHashMap<>();
+
+    public static boolean obtainDeliveryAgreement(CloseableHttpClient httpClient, String targetUrl, String requestOrigin) {
+        logger.info("obtain webhook delivery agreement for url: {}", targetUrl);
+        HttpOptions builder = new HttpOptions(targetUrl);
+        builder.addHeader(REQUEST_ORIGIN_HEADER, requestOrigin);
+
+        try (CloseableHttpResponse response = httpClient.execute(builder)) {
+            String allowedOrigin = response.getLastHeader(ALLOWED_ORIGIN_HEADER).getValue();
+            return allowedOrigin.equals("*") || allowedOrigin.equalsIgnoreCase(requestOrigin);
```
ruanwenjun, line 58: Can `allowedOrigin` be null? There will throw NPE if allowedOrigin is null.

eventmesh-common/src/main/java/org/apache/eventmesh/common/Constants.java
```diff
@@ -35,7 +35,7 @@
 
     public static final String PROTOCOL_DESC = "protocoldesc";
 
-    public static final int DEFAULT_HTTP_TIME_OUT = 3000;
+    public static final int DEFAULT_HTTP_TIME_OUT = 15000;
```
ruanwenjun, line 38: Why change this value? It's better to add comment(This is a history issue, but please add comment here)

jinrongluo: CloudEvents Webhook spec requires to Subscriber to implement Abuse Protection.  It means during the subscription, EventMesh needs to submit HTTP OPTIONS request to obtain the Delivery Consent . 

That means the HTTP subscription request is taking a bit longer than 3000ms.

So I increase the DEFAULT_HTTP_TIME_OUT

ruanwenjun: Ok, so just in `Webhook` we need to increase this timeout bigger than 3000?

It might be better to add a new Variable `DEFAULT_HTTP_WEBHOOK_TIME_OUT = 15000`, wdyt.

xwm1992 (dismissed):

xwm1992 (approved):

Issue #780, opened 2022-02-22
[Enhancement] EventListener defines in the level of Consumer may be more proper instead of  the level of topic


lrhkobe:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/eventmesh/issues?q=is%3Aissue) and found no similar issues.

### Enhancement Request

The eventListner in `Consumer.java` may be more proper if defined in the level of Consumer,instead of the level of Topic. Because the logic of eventListener is same for each topic in EventMesh.

```
void subscribe(String topic, final EventListener listener);
```

### Describe the solution you'd like

- modify method `subscribe` in `Consumer.java`, delete the binding between topic and eventListener
```
  //void subscribe(String topic, final EventListener listener) throws Exception;

    void subscribe(String topic) throws Exception;
```
- add method `registerEventListener` in `Consumer.java` , regist eventListener and store it in the level of Consumer

```
    void registerEventListener(EventListener listener);
```

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #781, opened 2022-02-22
[Issue #780] Modify the define level  of EventListener from Topic to Consumer

lrhkobe:
### Motivation

The logic of EventListner is same for each topic in EventMesh, and if new  an object of EventListner in heap for each topic, may lead to memory waste.

### Modifications

- modify method `subscribe` in `Consumer.java`, delete the binding between topic and eventListener
```
  //void subscribe(String topic, final EventListener listener) throws Exception;

    void subscribe(String topic) throws Exception;
```
- add method `registerEventListener` in `Consumer.java` , regist eventListener and store it in the level of Consumer
```
    void registerEventListener(EventListener listener);
```


xwm1992 (approved):
LGTM

Issue #614, opened 2021-11-25
[Enhancement] Add openschema dependency to eventmesh


yzhao244:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/eventmesh/issues?q=is%3Aissue) and found no similar issues.

### Enhancement Request

I would like to suggest to add openschema dependency to eventmesh project and create a client to communicate with openschema registry. Since openschema community has recently published the first version of OpenSchema Registry implementation which is a standalone schema registry server and maintained by the OpenSchema community,  eventmesh project can directly make a reference to the openschema dependency instead of maintaining openschema implemention within eventmesh project itself. 

### Describe the solution you'd like

Remove the current openschema implementation out of eventmesh project and add openschema dependency to eventmesh project. 

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

yzhao244:
@ruanwenjun @qqeasonchen

qqeasonchen:
ok, do we need confirm with @JunjieChou ?

JunjieChou:
> ok, do we need confirm with @JunjieChou ?

sure

JunjieChou:
> ok, do we need confirm with @JunjieChou ?

sure ok

vongosling:
Any update for this thread?

qqeasonchen:
@yzhao244 hi, Could you please help to fix it?

jzhou59:
Hi, I have a pr for the schema client api. I could continue this.

PR #821, opened 2022-04-03
[ISSUE #614] Add schema plugin and its api

jzhou59:
Fixes ISSUE #614 .

### Motivation

This pr adds a plugin for schema and adds the api for it.

### Documentation
This pr introduces a new feature schema client. 
The general schema design is presented in the repo, but the client is not documented yet.
I will update the schema documentation later.


eventmesh-schema-plugin/eventmesh-schema-api/src/main/java/org/apache/eventmesh/schema/api/SchemaPluginFactory.java
```diff
@@ -0,0 +1,40 @@
...
+     * Get {@code SchemaRegistry}
+     *
+     * @param schemaRegistryType the name of schema registry that configured in properties file
+     * @return schema registry plugin or null
+     */
+    public static SchemaRegistry getSchemaRegistry(String schemaRegistryType) {
+        checkNotNull(schemaRegistryType, "SchemaRegistryType cannot be null");
+        SchemaRegistry schemaRegistry = EventMeshExtensionFactory.getExtension(SchemaRegistry.class, schemaRegistryType);
+        return checkNotNull(schemaRegistry, "MetricsRegistryType: " + schemaRegistryType + " is not supported");
+    }
```
ruanwenjun, line 39:
```suggestion
        return checkNotNull(schemaRegistry, "SchemaRegistryType: " + schemaRegistryType + " is not supported");
```

eventmesh-schema-plugin/eventmesh-schema-api/src/main/java/org/apache/eventmesh/schema/api/SchemaType.java
```diff
@@ -0,0 +1,36 @@
...
+public enum SchemaType {
+    JSON("JSON"),
+    AVRO("AVRO"),
+    PROTOBUF("PROTOBUF");
+
+    private final String schemaTypeName;
+
+    SchemaType(String schemaTypeName) {
+        this.schemaTypeName = StringUtils.upperCase(StringUtils.deleteWhitespace(schemaTypeName));
+    }
```
ruanwenjun, line 31: Use `this.schemaTypeName = schemaTypeName` is better.

ruanwenjun (changes requested):
You may need to add `gradle.properties`, this will be used when packaging.

BTY, I think your `SchemaRegistry` design is not great. I suggest the schema just need to considerate how to serialize and deserialize. It doesn't need to take care about how to register schema, this should be implemen at `RegisterCenter` plugin.

eventmesh-schema-plugin/eventmesh-schema-api/src/main/java/org/apache/eventmesh/schema/api/SchemaPluginFactory.java
jzhou59: fixed

eventmesh-schema-plugin/eventmesh-schema-api/src/main/java/org/apache/eventmesh/schema/api/SchemaType.java
jzhou59: Ok, I was trying to specify all schema types to uppercase.
But you are right, I have altered the schema type according to IETF(or official) abbreviation.

References:
- [JSON](https://datatracker.ietf.org/doc/html/rfc8259)
- [Avro](https://avro.apache.org/docs/current/spec.html)
- [Protobuf](https://datatracker.ietf.org/doc/html/draft-rfernando-protocol-buffers-00#section-1.1)

jzhou59:
> You may need to add `gradle.properties`, this will be used when packaging.
> 
> BTY, I think your `SchemaRegistry` design is not great. I suggest the schema just need to considerate how to serialize and deserialize. It doesn't need to take care about how to register schema, this should be implemen at `RegisterCenter` plugin.

Sure, currently it is only ```api``` package and I will add ```gradle.properties``` in implementations.

The registering-related and retrieving-related procedures are indeed better in ```registry```-related plugin and I will modify it later.

However, I am not very clear about what should Eventmesh do with schema.

As far as I am concerned, there are four main concepts of schema: contract, serialization, deserialization, and validity.
- For contract, a schema is a contracted format of message between producer and consumer.
- For serialization, a producer serializes a message according to the contracted schema.
- For deserialization, a consumer deserializes a message according to the contracted schema.
- For validity, usually, a consumer checks the validity of the inbound serialized message before deserializing it.

So I was thinking that Eventmesh could take care of validity. If the inbound message from the producer is not obeying the specified schema, Eventmesh will discard this message and send an error status to the producer. The corresponding modules could be ```eventmesh-schema-json ```, ```eventmesh-schema-avro ```, and ```eventmesh-schema-protobuf ```.

Also, I get your point. Thinking from object-oriented programming, schema plugins could also provide the functions of ```serialize``` and ```deserialize```. Correct me if I had a misunderstanding. 


What do you think?

ruanwenjun:
> 

You are right, the api of schema plugin should provide the 4 interface `contract, serialization, deserialization, and validity `.
I think it shouldn't take care about how to store the schema(like register schema).

jzhou59:
> > 
> 
> Yes, you are right, the api of schema plugin should provide the 4 interface `contract, serialization, deserialization, and validity `. I think it shouldn't take care about how to store the schema(like register schema).

Ok, I have removed the schema storage-related functions from the schema plugin. What do you think of the current schema plugin structure of design?

Next, I will implement three examples of schema: JSON, Avro, and Protobuf respectively.

Also, where do you think the schema storage-related functions should be put? Maybe newly create a ```eventmesh-schemaregistry-plugin```?

ruanwenjun:
> 

If the schema should be stored, it can be seen as a metadata for topic, so can be implemented i metadata store module.

jzhou59:
Oh, not only storage, I mean the interaction between Eventmesh and openschemaregistry(such as retrieving, caching, and registering). Where should they be put?

ruanwenjun:
> Oh, not only storage, I mean the interaction between Eventmesh and schemaregistry(such as retrieving, caching, and registering). Here schemaregistry is different from schema. Since schema-plugin is only designed as schema-related. Where should schemaregistry be put?

The schema is seen as the metadata of a topic. When you register a schema to a topic, you just bind the schema to the topic, so you don't need to add a `SchemaRegistey`, we will have many methods to modify the metadata, we will a Metadata module to manage these data.

PR #848, opened 2022-04-29
[ISSUE #847] Could not find or load main class in eventmesh-examples

### LINKED ISSUE #847: [Bug] Could not find or load main class in eventmesh-examples
### Search before asking

### Environment

Linux

### EventMesh version

1.4.0

### What happened

Error: Could not find or load main class org.apache.eventmesh.tcp.demo.AsyncSubscribe

### How to reproduce

sh http_pub.sh
sh tcp_pub.sh
...

### Debug logs

Error: Could not find or load main class org.apache.eventmesh.tcp.demo.AsyncSubscribe

xuesongxs:
fix #847


ruanwenjun (changes requested):
Thanks for your patch, since we have an example for each protocol. 
So it's better to rename the script name or add an extra arg for script to choose which protocol example to run.

Issue #823, opened 2022-04-06
[Bug] upgrade spring libs because current version brings in jars with CVEs


xwm1992:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/eventmesh/issues?q=is%3Aissue) and found no similar issues.

### Environment

Window

### EventMesh version

master

### What happened

A Spring MVC or Spring WebFlux application running on JDK 9+ may be vulnerable to remote code execution (RCE) via data binding. The specific exploit requires the application to run on Tomcat as a WAR deployment. If the application is deployed as a Spring Boot executable jar, i.e. the default, it is not vulnerable to the exploit. However, the nature of the vulnerability is more general, and there may be other ways to exploit it.

https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2022-22965

`spring-boot-starter-web` should update to version 2.6.6 

### How to reproduce

N/A

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

Issue #822, opened 2022-04-06
[Bug] upgrade jackson-databind because current version brings the CVEs


xwm1992:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/eventmesh/issues?q=is%3Aissue) and found no similar issues.

### Environment

Windows

### EventMesh version

master

### What happened

jackson-databind before 2.13.0 allows a Java StackOverflow exception and denial of service via a large depth of nested objects.

https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-36518

`jackson-databind` version should update to version 2.13.0

### How to reproduce

N/A

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

PR #849, opened 2022-05-05
upgrade spring libs & jackson-databind s version to fix CVEs issues

misselvexu:
### Contribution Checklist

Fixes ISSUE #822 .
Fixes ISSUE #823  .

### Motivation

- Upgrade `spring libs` because current version brings in jars with CVEs.
- Upgrade `jackson-databind` because current version brings the CVEs

### Modifications

Edit `build.gradle` file.

```
// Jackson Dependency
dependency "com.fasterxml.jackson.core:jackson-databind:2.13.0"
dependency "com.fasterxml.jackson.core:jackson-core:2.13.0"
dependency "com.fasterxml.jackson.core:jackson-annotations:2.13.0"

// Spring Boot Web Dependency
dependency "org.springframework.boot:spring-boot-starter-web:2.6.7"
```

### Documentation

none.


li-xiao-shuang (approved):

xwm1992:
image
please update the upgrade dependencies version in these files.

misselvexu:
> image please update the upgrade dependencies version in these files.

Got it .

tools/third-party-dependencies/known-dependencies.txt
```diff
@@ -130,4 +130,5 @@ httpcore-nio-4.4.6.jar
 javassist-3.21.0-GA.jar
 nacos-client-2.0.4.jar
 reflections-0.9.11.jar
-snakeyaml-1.23.jar
\ No newline at end of file
+snakeyaml-1.23.jar
+spring-boot-starter-web-2.6.7.jar
```
ruanwenjun, line 134: Right now, the example module will not be packaged into distribution, so we may don't need to add license of `spring-boot-starter-web-2.6.7`? Actually, I suggest we add example into our distribution. cc @xwm1992

ruanwenjun (changes requested):

tools/third-party-dependencies/known-dependencies.txt
misselvexu: So ? You decide .

xwm1992: > Right now, the example module will not be packaged into distribution, so we may don't need to add license of `spring-boot-starter-web-2.6.7`? Actually, I suggest we add example into our distribution. cc @xwm1992

not only for distribution, theses modules are all upload to the maven repository.

xwm1992: > Currently, the check license script will only check the third-part libs in distribution. If we didn't add the example module in distribution, we can't add `spring-boot-starter-web-2.6.7.jar ` in this file, otherwise the license check will fail.

Ok, I get what you mean,  we either only upgrade the `spring-boot-starter-web-2.6.7.jar` version and not add this to the file `known-dependencies.txt`, or add the `example moudule` into the distribution ? I prefer the former, not add  `spring-boot-starter-web-2.6.7.jar` to the files, because the `example` don't need to be used in distributions. @misselvexu @ruanwenjun what's your opinion?

misselvexu: Already removed temporarily .

ruanwenjun: In this PR, we don't need to add `pring-boot-starter-web-2.6.7.jar` into `known-dependencies.txt `. We can add another PR to add example module into distribution.

ruanwenjun: Before submit PR, you can discuss this in issue or mail, I am not sure if this is needed.

Issue #856, opened 2022-05-10
[Enhancement] Add test code for this module [eventmesh-trace-plugin]


Alonexc:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/eventmesh/issues?q=is%3Aissue) and found no similar issues.

### Enhancement Request

Add test code for eventmesh-trace-plugin module.

### Describe the solution you'd like

module name: eventmesh-trace-api and eventmesh-trace-zipkin
located in: eventmesh-trace-plugin
Task status: ×(unfinished) / √(finished)

> Notes: This issue will not be closed, if you want to submit a pr to do some unit tests on this module or select some of the following classes to do some unit tests, please submit a detailed issue separately and linked to the pr you submitted.

| Task Status | Class | Type|
| :------: | :------| :------|
| × | org/apache/eventmesh/trace/api/config/ExporterConfiguration.java | config |
| × | org/apache/eventmesh/trace/api/TracePluginFactory.java | service |
| × | org/apache/eventmesh/trace/zipkin/config/ZipkinConfiguration.java | config |

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

HoffmanZheng:
I'd like to work on this. plz assign this to me.

Alonexc:
@HoffmanZheng Hello! This issue will not be deleted,you can provide another detailed issue for your unit test and associate it with the pr you submitted.

HoffmanZheng:
> @HoffmanZheng Hello! This issue will not be deleted,you can provide another detailed issue for your unit test and associate it with the pr you submitted.

little confused here, why do I need another detailed issue ? To document the test code i write? IMO, the test code is already human readable.

Alonexc:
@HoffmanZheng Because an issue will be closed when it is finished, and this issue will not be closed, the unit test module should be able to continuously supplement, so you need to mention another issue.

HoffmanZheng:
A detailed issue #871 added.

Pil0tXia:
Completed in https://github.com/apache/eventmesh/issues/871.

PR #867, opened 2022-05-15
[ISSUE-856] Add test code for module [eventmesh-trace-plugin]

HoffmanZheng:
Fixes ISSUE #856.

### Motivation

Add test code for eventmesh-trace-plugin module.

### Modifications

Add unit test for `org/apache/eventmesh/trace/api/TracePluginFactory.java`.

Add unit test for `org/apache/eventmesh/trace/zipkin/config/ZipkinConfiguration.java`.

### Documentation

Not a new feature, thus no doc needed.

### DIFF
--- a/eventmesh-trace-plugin/eventmesh-trace-zipkin/src/main/java/org/apache/eventmesh/trace/zipkin/ZipkinTraceService.java
+++ b/eventmesh-trace-plugin/eventmesh-trace-zipkin/src/main/java/org/apache/eventmesh/trace/zipkin/ZipkinTraceService.java
@@ -56,6 +56,8 @@ public class ZipkinTraceService implements TraceService {
 
     private OpenTelemetry openTelemetry;
 
+    private Thread shutdownHook;
+
     @Override
     public void init() {
         //zipkin's config
@@ -92,7 +94,8 @@ public class ZipkinTraceService implements TraceService {
             .setTracerProvider(sdkTracerProvider)
             .build();
 
-        Runtime.getRuntime().addShutdownHook(new Thread(sdkTracerProvider::close));
+        shutdownHook = new Thread(sdkTracerProvider::close);
+        Runtime.getRuntime().addShutdownHook(shutdownHook);
     }
 
     @Override
--- a/eventmesh-trace-plugin/eventmesh-trace-zipkin/src/test/java/org/apache/eventmesh/trace/zipkin/ZipkinTraceServiceTest.java
+++ b/eventmesh-trace-plugin/eventmesh-trace-zipkin/src/test/java/org/apache/eventmesh/trace/zipkin/ZipkinTraceServiceTest.java
@@ -1,0 +1,74 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.eventmesh.trace.zipkin;
+
+import static org.junit.Assert.*;
+
+import java.lang.reflect.Field;
+
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.Mockito;
+import org.mockito.junit.MockitoJUnitRunner;
+
+import io.opentelemetry.api.OpenTelemetry;
+import io.opentelemetry.sdk.trace.SdkTracerProvider;
+
+@RunWith(MockitoJUnitRunner.class)
+public class ZipkinTraceServiceTest {
+
+    @Test
+    public void init() throws NoSuchFieldException, IllegalAccessException {
+        ZipkinTraceService zipkinTraceService = new ZipkinTraceService();
+        zipkinTraceService.init();
+
+        Field sdkTracerProviderField = ZipkinTraceService.class.getDeclaredField("sdkTracerProvider");
+        Field openTelemetryField = ZipkinTraceService.class.getDeclaredField("openTelemetry");
+        Field shutdownHookField = ZipkinTraceService.class.getDeclaredField("shutdownHook");
+        sdkTracerProviderField.setAccessible(true);
+        openTelemetryField.setAccessible(true);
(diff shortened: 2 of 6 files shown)


xiaoyang-sde:
Welcome and thanks for your contribution!

HoffmanZheng:
There is no exposed behavior in [org/apache/eventmesh/trace/api/config/ExporterConfiguration.java](https://github.com/apache/incubator-eventmesh/blob/master/eventmesh-trace-plugin/eventmesh-trace-api/src/main/java/org/apache/eventmesh/trace/api/config/ExporterConfiguration.java).
Not sure whether it is necessary to make `initializeConfig()` and `loadProperties()` protected (package private), and add test for them.

xwm1992:
`eventmesh-trace-plugin/eventmesh-trace-zipkin/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker`, @HoffmanZheng hi, you need add the apache license header for this file.

HoffmanZheng:
> `eventmesh-trace-plugin/eventmesh-trace-zipkin/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker`, @HoffmanZheng hi, you need add the apache license header for this file.

Once I add the license head, the Mockito test failed. As I mentioned above, this file might be single line to enable the incubating feature of Mockito (for our situation, to mock the final class).

HoffmanZheng:
> > > `eventmesh-trace-plugin/eventmesh-trace-zipkin/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker`, @HoffmanZheng hi, you need add the apache license header for this file.
> > 
> > 
> > Once I add the license head, the Mockito test failed. As I mentioned above, this file might be **single line** to enable the incubating feature of Mockito (for our situation, to mock the final class). 
> 
> If so, I suggest that we can use `PowerMock` to mock the final class, since we must contain the license header in our source file.

didn't find the solution how PowerMock mocks the final class, but found this to solve the problem.
https://frontbackend.com/java/mocking-final-method-with-mockito-framework

eventmesh-trace-plugin/eventmesh-trace-zipkin/src/main/java/org/apache/eventmesh/trace/zipkin/ZipkinTraceService.java
```diff
@@ -92,7 +94,8 @@ public void init() {
             .setTracerProvider(sdkTracerProvider)
             .build();
 
-        Runtime.getRuntime().addShutdownHook(new Thread(sdkTracerProvider::close));
+        shutdownHook = new Thread(sdkTracerProvider::close);
+        Runtime.getRuntime().addShutdownHook(shutdownHook);
```
ruanwenjun, line 98: I don't think this change is needed.

eventmesh-trace-plugin/eventmesh-trace-zipkin/src/test/java/org/apache/eventmesh/trace/zipkin/ZipkinTraceServiceTest.java
```diff
@@ -0,0 +1,70 @@
...
+    public void testInit() throws NoSuchFieldException, IllegalAccessException {
+        ZipkinTraceService zipkinTraceService = new ZipkinTraceService();
+        zipkinTraceService.init();
+
+        Field sdkTracerProviderField = ZipkinTraceService.class.getDeclaredField("sdkTracerProvider");
+        Field openTelemetryField = ZipkinTraceService.class.getDeclaredField("openTelemetry");
+        Field shutdownHookField = ZipkinTraceService.class.getDeclaredField("shutdownHook");
+        sdkTracerProviderField.setAccessible(true);
+        openTelemetryField.setAccessible(true);
+        shutdownHookField.setAccessible(true);
```
ruanwenjun, line 43: The better way is to change the visibility in `ZipkinTraceService` rather than use reflect.

ruanwenjun (changes requested):

eventmesh-trace-plugin/eventmesh-trace-zipkin/src/main/java/org/apache/eventmesh/trace/zipkin/ZipkinTraceService.java
HoffmanZheng: ^ to keep the reference to the shutdownHookThread, thus the behavior of `addShutdownHook` could be tested (it raise `Hook previously registered` when the same shutdownHookThread is added at second time).

https://github.com/apache/incubator-eventmesh/blob/6ea8944b8c91314b296acbb9ef107902c32b084d/eventmesh-trace-plugin/eventmesh-trace-zipkin/src/test/java/org/apache/eventmesh/trace/zipkin/ZipkinTraceServiceTest.java#L47-L55

the `new Thread(sdkTracerProvider::close)` is not the same shutdownHookThread as previous, it could be added successfully and doesn't raise any exception (thus the behavior of `addShutdownHook` could not be tested).

https://github.com/apache/incubator-eventmesh/blob/836e3a5bd6a25c9831b0c6a8db30c997c68f0e91/eventmesh-trace-plugin/eventmesh-trace-zipkin/src/main/java/org/apache/eventmesh/trace/zipkin/ZipkinTraceService.java#L95

ruanwenjun: In your assert, you add a shutdown hook twice, then it throws an exception. Does this test make sense?

HoffmanZheng: > add a shutdown hook twice, then it throws an exception

the shutdown hook is already added in the `zipkinTraceService.init();`, thus it throws exception in test as expected.

It ensures the `addShutdownHook` is called in the `init()` method, and the `Thread(sdkTracerProvider::close)` is added there. If someone change the behavior of that (e.g. remove the whole line `Runtime.getRuntime().addShutdownHook(new Thread(sdkTracerProvider::close));`), the unit test would fail.

ruanwenjun: OK, I got it.

eventmesh-trace-plugin/eventmesh-trace-zipkin/src/test/java/org/apache/eventmesh/trace/zipkin/ZipkinTraceServiceTest.java
HoffmanZheng: Do you mean change the modifier of these member variable from private to `protected` ?
That makes sense to me, the test code located in the same package of `ZipkinTraceService `, which makes the protected member variable also accessible.

ruanwenjun: Yes.

ruanwenjun (approved):
LGTM, thanks for your patch.

Issue #883, opened 2022-05-21
[Enhancement] Transitively export the dependencies of [eventmesh-sdk-java] to other module at compile time 


HoffmanZheng:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/eventmesh/issues?q=is%3Aissue) and found no similar issues.

### Enhancement Request

Issue found when I'm working on #579, and trying to run a demo application with eventmesh-sdk-java.
Some dependencies of [Eventmesh-sdk-java] are in **runtime** scope, which are also necessary to build an application with that library. It is not convenient for developers and lib module distribution.
![1653116211583](https://user-images.githubusercontent.com/45369893/169639930-a23c51a4-ec36-4c53-a04b-978d1cbde210.jpg)

### Describe the solution you'd like

Change the dependencies configuration in [Eventmesh-sdk-java], which are needed in the compile phase.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #884, opened 2022-05-21
[ISSUE-883] Transitively export the dependencies of [eventmesh-sdk-java] to other module

HoffmanZheng:
Fixes ISSUE #883 .

### Motivation

[eventmesh-sdj-java](https://mvnrepository.com/artifact/org.apache.eventmesh/eventmesh-sdk-java) is a library module, should transitively export the dependencies, which are necessary for development of [eventmesh-application](https://github.com/apache/incubator-eventmesh/tree/master/eventmesh-examples).

### Modifications

Add gradle plugin [java-library](https://docs.gradle.org/current/userguide/java_library_plugin.html) in eventmesh-sdk-java

Change the configuration of following dependencies from `implementation` to `api`: 
* project(":eventmesh-common")
* io.cloudevents:cloudevents-core
* io.cloudevents:cloudevents-json-jackson
* io.openmessaging:openmessaging-api
* org.projectlombok:lombok:1.18.22

### Documentation

Not a new feature, thus no doc needed.


HoffmanZheng (commented):
Though [maven-publish](https://github.com/apache/incubator-eventmesh/blob/53c62509012dafa4d0912f9f7cb35d5be6c11389/build.gradle#L52) is included in the project, there should be a committor to do the related test.

eventmesh-sdk-java/build.gradle
```diff
@@ -15,10 +15,15 @@
  * limitations under the License.
  */
 
+plugins {
+    id 'java-library'
+    id 'com.github.johnrengelman.shadow' version '7.1.2'
```
HoffmanZheng, line 20: After I use the `java-library` and change the dependency configuration to `api`, the built jar file still does not contain the dependencies.
Thus the `shadow` plugin used here to generate a fat jar. (for your [reference](https://discuss.gradle.org/t/how-to-include-dependencies-in-jar/19571))

HoffmanZheng (commented):
Though [maven-publish](https://github.com/apache/incubator-eventmesh/blob/53c62509012dafa4d0912f9f7cb35d5be6c11389/build.gradle#L52) is included in the project, there should be a committor to do the related test.

ruanwenjun:
Good catch. Right now, the scope of `eventmesh-common` in `eventmesh-sdk-java` is `implementation`, so if we want to use eventmesh-sdk-java in our app, we need to add the `eventmesh-common` to our dependency. We need to change the scope from `implementation` to api to solve this.

And I think we don't need to make a shawdow jar. cc @xwm1992

HoffmanZheng:
> Good catch. Right now, the scope of `eventmesh-common` in `eventmesh-sdk-java` is `implementation`, so if we want to use eventmesh-sdk-java in our app, we need to add the `eventmesh-common` to our dependency. We need to change the scope from `implementation` to api to solve this.
> 
> And I think we don't need to make a shawdow jar. cc @xwm1992

Currently I'm trying to use the following script to generate a fat jar of [eventmesh-sdk-java], sometimes it works. Mark the PR work in progress.

```
jar {
    from {
        configurations. runtimeClasspath.filter{ it.exists() }.collect { it.isDirectory() ? it : zipTree(it) }
    }
}
```

ruanwenjun:
> > Good catch. Right now, the scope of `eventmesh-common` in `eventmesh-sdk-java` is `implementation`, so if we want to use eventmesh-sdk-java in our app, we need to add the `eventmesh-common` to our dependency. We need to change the scope from `implementation` to api to solve this.
> > And I think we don't need to make a shawdow jar. cc @xwm1992
> 
> Currently I'm trying to use the following script to generate a fat jar of [eventmesh-sdk-java] (without the usage of `shadow` plugin), sometimes it works. Mark the PR work in progress.
> 
> ```
> jar {
>     from {
>         configurations. runtimeClasspath.filter{ it.exists() }.collect { it.isDirectory() ? it : zipTree(it) }
>     }
> }
> ```

I suggest we just change the scope of `eventmesh-commom` to `spi` is enough.

HoffmanZheng:
I used to use the `gradle jar` and `mvn install:install-file` to do dependencies resolving test, and it does not work.
Rather using `gradle publishToMavenLocal`, the dependencies could be transitively exported to other module!

eventmesh-sdk-java/build.gradle
```diff
@@ -15,10 +15,14 @@
  * limitations under the License.
  */
 
+plugins {
+    id 'java-library'
+}
+
```
ruanwenjun, line 21: It seems this already defined at root `build.gradle`
```suggestion
```

ruanwenjun (changes requested):

qqeasonchen:
@HoffmanZheng any update？

eventmesh-sdk-java/build.gradle
```diff
@@ -50,10 +54,9 @@ dependencies {
     implementation "io.grpc:grpc-protobuf:1.15.0"
     implementation "io.grpc:grpc-stub:1.15.0"
     implementation "com.google.protobuf:protobuf-java-util:3.5.1"
-    compileOnly 'org.projectlombok:lombok:1.18.22'
+    api 'org.projectlombok:lombok:1.18.22'
```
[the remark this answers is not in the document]
HoffmanZheng: also make sense to me, seems the `eventmesh-example` uses that only for log. it's not necessary to export to the users of `eventmesh-sdk-java`
https://github.com/apache/incubator-eventmesh/blob/c055e8324fd6b0c67055b2f4c6731f386acdb5a4/eventmesh-examples/src/main/java/org/apache/eventmesh/http/demo/sub/controller/SubController.java#L42
https://github.com/apache/incubator-eventmesh/blob/c055e8324fd6b0c67055b2f4c6731f386acdb5a4/eventmesh-examples/src/main/java/org/apache/eventmesh/http/demo/sub/controller/SubController.java#L56

HoffmanZheng:
> @HoffmanZheng any update？

updated

ruanwenjun (approved):
LGTM

PR #912, opened 2022-06-07
[Go SDK] Update licenserrc.yaml to ignore go.sum

walleliu1016:
- update licenserrc.yaml ignore go.sum
- add license header for some go files


xwm1992 (approved):
LGTM

Issue #923, opened 2022-06-17
[Bug] The key is not right when remove from the map `ClientGroupMap`  in `ClientSessionGroupMapping.java`


lrhkobe:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/eventmesh/issues?q=is%3Aissue) and found no similar issues.

### Environment

Windows

### EventMesh version

master

### What happened

- when put item in `ClientGroupMap`

![image](https://user-images.githubusercontent.com/34571087/174215089-4dc8101d-68c0-43b1-a4bf-d67654b30ce7.png)

- when remove item in `ClientGroupMap`

![image](https://user-images.githubusercontent.com/34571087/174215294-3d9eac2b-d22a-48d4-a705-20315d780128.png)

### How to reproduce

Modify the removed key same with when putted.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #949, opened 2022-06-22
[ISSUE #923] Fix the key when removing item from ClientGroupMap

lrhkobe:
Fixes ISSUE #923 

### Motivation

- Removing the ClientGroupWrapper from map not work, because the key is not right.

### Modifications

- Modify the removed key same with when putted.


xwm1992 (approved):
LGTM

Issue #946, opened 2022-06-22
[Enhancement] trace optimize


lrhkobe:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/eventmesh/issues?q=is%3Aissue) and found no similar issues.

### Enhancement Request

- Add trace buried point in eventmesh-runtime
- Record specific attributes to the trace, for example, the time of the event life cycle, the eventmesh ip that the event passed through

### Describe the solution you'd like

- Adjust eventmesh-trace-api，which is based on opentelemetry-api
- Provide a more specific trace implementation，add trace buried point in eventmesh-runtime

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #952, opened 2022-06-22
[ISSUE #946] Optimize trace module in eventmesh

lrhkobe:
Fixes ISSUE #946 .

### Motivation

- The current trace implementation is relatively simple ，may not general enough for other specific trace plugin implementations.

### Modifications

- Adjust eventmesh-trace-api，which is based on opentelemetry-api
```
public interface EventMeshTraceService {
    void init() throws TraceException;

    //extract attr from carrier to context
    Context extractFrom(Context context, Map<String, Object> carrier) throws TraceException;

    //inject attr from context to carrier
    void inject(Context context, Map<String, Object> carrier);

    Span createSpan(String spanName, SpanKind spanKind, long startTimestamp, TimeUnit timeUnit,
                    Context context, boolean isSpanFinishInOtherThread) throws TraceException;

    Span createSpan(String spanName, SpanKind spanKind, Context context,
                    boolean isSpanFinishInOtherThread) throws TraceException;

    void shutdown() throws TraceException;
}
```


xwm1992 (approved):
LGTM

PR #963, opened 2022-06-24
[ISSUE #957] add trace for http protocol in eventmesh-runtime

lrhkobe:
Fixes ISSUE #957 .

### Motivation

- In cloud-native application scenarios, the flow of events is very complex and common, event tracking is very meaningful and important for analyzing problems

### Modifications

- Add trace buried point for http protocol in eventmesh-runtime
- Record specific attributes to the trace, for example, the time of the event life cycle, the eventmesh ip that the event passed through


xwm1992 (approved):
LGTM

Issue #970, opened 2022-06-28
[Enhancement] Include lombok for every submodule, same as junit


wuxuanqicn:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/eventmesh/issues?q=is%3Aissue) and found no similar issues.

### Enhancement Request

As a basic component, we can include lombok for every submodule by default

### Describe the solution you'd like

build.gradle 
```
allprojects {
    dependencies {
        testImplementation "junit:junit:4.13.2"

        compileOnly 'org.projectlombok:lombok'
        annotationProcessor 'org.projectlombok:lombok'

        testCompileOnly 'org.projectlombok:lombok'
        testAnnotationProcessor 'org.projectlombok:lombok'
    }
}
```

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #971, opened 2022-06-28
include lombok for every submodule by default

wuxuanqicn:
Fixes ISSUE #970

### Motivation
As a basic component, we can include lombok for every submodule by default

### Modifications
1.add dependency at root build.gradle file
2. remove submodule lombok dependency

### Documentation
dependency change


wuxuanqicn:
close due to no reviewer

PR #972, opened 2022-06-28
Update README.md

qqeasonchen:
update slack


li-xiao-shuang (approved):

ruanwenjun (approved):
LGTM. I think we can let user send their request to dev@eventmesh.apache.org when they want to join the slack, rather than we need to update the link every month.

PR #991, opened 2022-07-04
Fix Error when Running/Debugging the Application on Linux Platform

Markliniubility:
### Motivation

When the Eventmesh application runs on a Linux Platform using IDE, there will be errors on reading the properties and setting files. The error was due to the mix-use of forward-slash and backward-slash. The bug impacts developers who use IDE on Mac or Linux Platform

### Modifications

Replace all the / and \ with proper file separators that work both for Linux and Windows

### Documentation

- No
- No new feature introduced


Alonexc (approved):

xiaoyang-sde (approved):

xiaoyang-sde (commented):
Will this change breaks the code in Windows platform?

Markliniubility:
> Will this change breaks the code in Windows platform?

No, it doesn't., as I have tested, and the Github Action tests have shown

xiaoyang-sde (approved):
LGTM

PR #1020, opened 2022-07-13
[ISSUE #1021]Fix the case that span is null

liuzwei:
Fixed ISSUE #1021 

### Motivation

When I run the `eventmesh-runtime` in debug model. I find a bug which is NullpointException at line 432 in file `ClientGroupWrapper ` . So I want to  Fix it.

### Modifications

I read some information about the telemetry. I think the method `createSpan` in file `Trace` should not return `null`, at least return a invalid Span .

### Documentation

- Does this pull request introduce a new feature? 
- no


Alonexc (commented):
Hi, you need to add an issue for this pr.

liuzwei:
> Hi, you need to add an issue for this pr.

OK

Alonexc:
Good job, you need to change the title to associate pr with issue.

xwm1992 (approved):

Issue #1016, opened 2022-07-12
[Enhancement] Implement 'NameResolver' and 'LoadBalancer' extension points on top of the Java SDK's protocol


mytang0:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/eventmesh/issues?q=is%3Aissue) and found no similar issues.

### Enhancement Request

Implement 'NameResolver' and 'LoadBalancer' extension points on top of the Java SDK's protocol. At present, the Java SDK lacks a unified load balancing extension, and only provides an optional load balancing implementation for the HTTP protocol  (see org.apache.eventmesh.common.loadbalance.LoadBalanceSelector).

### Describe the solution you'd like

Implement 'NameResolver' and 'LoadBalancer' extension points on top of the Java SDK's protocol, support user-defined load balancing implementation.

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

PR #974, opened 2022-06-29
[ISSUE #973] zookeeper registry

VNBear:
Impl ISSUE #973 

Motivation

And zookeeper registry

Modifications

And zookeeper registry

Documentation

Does this pull request introduce a new feature?  yes

PR #1030, opened 2022-07-19
[ISSUE #1016] Implement loadBalance for gRPC producer

VNBear:
Fixes ISSUE #1016 .

### Motivation

*java-sdk must be fixed service address*
*Expect to be able to get the service address automatically from the registry*

### Modifications

*wrapper java-sdk of grpc producer to impl loadbalance*

### Documentation

- Does this pull request introduce a new feature? (yes )


ruanwenjun:
> > I don't think it's time to do this change, we didn't register EventMesh runtime server in ZK.
> 
> I have previously submitted a PR support ZK registry #974 . My thinking is that the type of registry currently supported is not critical, it may be important to design the nameResolver and loadbalance of client well, subsequent registries can be expanded on this basis. This is just my personal opinion, can you give me some suggestions ?

You need to submit a mail to dev to discuss this, since this will make a great impact to EM. I agree we can add this feature, but you still need to send proposal(This is important), and we need to put the logic in your ZK registry, the client should only rely on registry to get/notify the change of server address.

VNBear:
> > > I don't think it's time to do this change, we didn't register EventMesh runtime server in ZK.
> > 
> > 
> > I have previously submitted a PR support ZK registry #974 . My thinking is that the type of registry currently supported is not critical, it may be important to design the nameResolver and loadbalance of client well, subsequent registries can be expanded on this basis. This is just my personal opinion, can you give me some suggestions ?
> 
> You need to submit a mail to dev to discuss this, since this will make a great impact to EM. I agree we can add this feature, but you still need to send proposal(This is important), and we need to put the logic in your ZK registry, the client should only rely on registry to get/notify the change of server address.

Alright. Thanks♪(･ω･)ﾉ

VNBear:
> > > I don't think it's time to do this change, we didn't register EventMesh runtime server in ZK.
> > 
> > 
> > I have previously submitted a PR support ZK registry #974 . My thinking is that the type of registry currently supported is not critical, it may be important to design the nameResolver and loadbalance of client well, subsequent registries can be expanded on this basis. This is just my personal opinion, can you give me some suggestions ?
> 
> You need to submit a mail to dev to discuss this, since this will make a great impact to EM. I agree we can add this feature, but you still need to send proposal(This is important), and we need to put the logic in your ZK registry, the client should only rely on registry to get/notify the change of server address.

Hi, I have sent the email some time ago. I wonder what is happening now.
image

VNBear:
> Hi, if possible, could you please split your pr into separate part? Smaller pr with fewer changed file would help us review your code better and faster. Hope you could understand, thank you for your contribution!

First of all, thank you for your advice. The ISSUE corresponding to this PR is a big change to the existing Client SDK, so it is submitted more. One of the members suggested sending an email to discuss this feature, so this PR is just a draft.

Issue #1043, opened 2022-07-27
[Bug] The problem of shared registry singleton


mytang0:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/eventmesh/issues?q=is%3Aissue) and found no similar issues.

### Environment

Linux

### EventMesh version

master

### What happened

1. registryService instance is overwritten on startup
2. unRegister does not complete normally

### How to reproduce

When there are more than two services at the same time, there will be problems with the start and shutdown of the registry. Services refer to TCPServer, HTTPServer and GRPCServer.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1044, opened 2022-07-27
[ISSUE #1043] Fix the problem of shared registry singleton.

mytang0:
Fixes ISSUE #1043

### DIFF
--- a/eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/registry/Registry.java
+++ b/eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/registry/Registry.java
@@ -31,23 +31,37 @@ import org.slf4j.LoggerFactory;
 
 public class Registry {
     private static final Logger logger = LoggerFactory.getLogger(Registry.class);
-    private static RegistryService registryService;
 
-    public void init(String registryPluginType) throws Exception {
-        registryService = EventMeshExtensionFactory.getExtension(RegistryService.class, registryPluginType);
-        if (registryService == null) {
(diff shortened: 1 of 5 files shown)


eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/registry/Registry.java
```diff
@@ -31,23 +31,37 @@
...
+            started = true;
+        }
     }
 
-    public void shutdown() throws Exception {
-        registryService.shutdown();
+    public synchronized void shutdown() throws Exception {
+        if (started) {
+            registryService.shutdown();
+            started = false;
```
li-xiao-shuang, line 63: inited = false;

li-xiao-shuang:
Register only one protocol service if three protocols are started at the same time?

li-xiao-shuang (approved):

eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/registry/Registry.java
mytang0: Thanks, fixed.

mytang0: > inited = false;

Thanks, fixed.

li-xiao-shuang (approved):

li-xiao-shuang:
LGTM

qqeasonchen:
@ruanwenjun please help to check,thanks.

Issue #1052, opened 2022-07-29
[Bug] Only the first instance of the same consumer group receives the message in http protocol


lrhkobe:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/eventmesh/issues?q=is%3Aissue) and found no similar issues.

### Environment

Windows

### EventMesh version

1.5.0

### What happened

When the same consumer group starts multiple listening instances, only the first started instance receives the message

### How to reproduce

1. Start an eventmesh instance
2. The same consumer group starts multiple listening instances, and the url (which is the parameter passed by the client when subscribing) of each instance is different
3. Send multiple event messages

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1053, opened 2022-07-29
[ISSUE #1052]fix push msg only the first started instance with same group

lrhkobe:
Fixes ISSUE #1052 .

### Modifications

When other instances of the same group subscribe, update the subscription information in `EventMeshConsumer`


xwm1992 (approved):
LGTM

Issue #1064, opened 2022-08-02
[Bug] NullPointException Of ClientManageControllerTest


walterlife:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/eventmesh/issues?q=is%3Aissue) and found no similar issues.

### Environment

Mac

### EventMesh version

master

### What happened

When running the ClientManageController unit test, NPE will appear. The reason is that due to the new Webhook feature, it is necessary to add related Mock adaptations.

### How to reproduce

add  AdminWebHookConfigOperationManage mock logic in ClientManageController unit test file.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1067, opened 2022-08-02
[ISSUE #1064]Fix NullPointException Of ClientManageControllerTest

walterlife:
Fixes ISSUE #1064 .

### Motivation
When running the ClientManageController unit test, NPE will appear. The reason is that due to the new Webhook feature, it is necessary to add related Mock adaptations.

### Modifications
add AdminWebHookConfigOperationManage mock logic in ClientManageController unit test file.

### Documentation
modify unit test,no need to add new documents


xwm1992 (approved):
LGTM

PR #1140, opened 2022-08-10
[ISSUE #1139] add plugin module

### LINKED ISSUE #1139: [Enhancement] eventmesh server go plugin
### Search before asking

### Enhancement Request

Complete the plugin module(include log,naming,database,config)

implement a pluggable architecture

walterlife:
Fixes #1139 .

### Motivation
add plugin module

### Modifications
1. the plugin module has been added to the eventmesh-go project, and log/naming/database has been implemented. 
2. replacing the existing log module, adding a rolllwrite component based on zap, providing better write performance, and asynchronously discarding logs when the disk is full, so that the service will not be blocked.
3. add server/main file to facilitate verification of new functions.

### Documentation


walleliu1016:
1. Move the code in internal into pkg package,  as eventmesh is an opensource project, code in the pkg package maybe better instead in the internal
2. Move the main.go into cmd/workflow/main.go as workflow is a single process, we will start another process in the cmd/runtime
3. I see you are using yaml configuration, then delete toml in the configs directory
4. Move the event-mesh-server.yaml into configs directory

walterlife:
> 1. Move the code in internal into pkg package,  as eventmesh is an opensource project, code in the pkg package maybe better instead in the internal
> 2. Move the main.go into cmd/workflow/main.go as workflow is a single process, we will start another process in the cmd/runtime
> 3. I see you are using yaml configuration, then delete toml in the configs directory
> 4. Move the event-mesh-server.yaml into configs directory

1. agree
2. It is not recommended to move the server.go and main.go files to the cmd directory. cmd is generally used to start programs based on the command line, and different configuration parameters can be passed in based on this. The server.go and main.go files here are the entry points for the eventmesh-server service. Personally, it is recommended to put them in the root directory or create a new server directory. Subsequent workflow/catalog will have a separate directory.
3/4 The current function of config package is to parse the configuration file when the service is running. If the configuration file is placed in this package, the responsibility of the package will be different (you can create a conf directory and put the configuration file in this directory when the service is running). ), the practice of etcd is also referred to here (the configuration file is placed directly in the root directory of the source code)

PR #1313, opened 2022-09-19
support protocol amqp

wangshaojie4039:
1. add remoting impl
2. add producer,consumer interface 
3. add connection ,channel interface


qqeasonchen:
@wangshaojie4039 good job~, but I suggest you split into 6-5 smaller prs ,thanks.

PR #1618, opened 2022-10-16
[ISSUE #1517]Update UnSubscribeResponseBody.java

### LINKED ISSUE #1517: [Enhancement] There is no need to declare exceptions [UnSubscribeResponseBody]
### Search before asking

### Enhancement Request

image
located at: org/apache/eventmesh/common/protocol/http/body/client/UnSubscribeResponseBody.java line 59
This method declares that it throws a checked exception that it does not throw. As this method is either a constructor, static method or private method, there is no reason for this method to declare the exception in its throws clause, and just causes calling methods to unnecessarily handle an exception that will never be thrown. The exception in question should be removed from the throws clause.

It is recommended to remove the content after throws.

jananip21:
Fixes #1517

### Motivation

There is no need to declare exceptions
The method is static so removing the throws exception

### Modifications

"throws exception" has been removed from the code

### Documentation

- Does this pull request introduce a new feature? No
- If yes, how is the feature documented? Not applicable
- If a feature is not applicable for documentation, explain why? No new feature is introduced

### DIFF
--- a/eventmesh-common/src/main/java/org/apache/eventmesh/common/protocol/http/body/client/UnSubscribeResponseBody.java
+++ b/eventmesh-common/src/main/java/org/apache/eventmesh/common/protocol/http/body/client/UnSubscribeResponseBody.java
@@ -56,7 +56,7 @@ public class UnSubscribeResponseBody extends Body {
         this.resTime = resTime;
     }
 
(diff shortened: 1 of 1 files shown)


githublaohu (approved):
LGTM

xwm1992 (approved):
LGTM

PR #1701, opened 2022-10-21
Fix. Remove throws exception.Fixed patch

rinuaby13:
ISSUE #1741
[Bug Fix]

Modifications
Fix patch in
 eventmesh-common/src/main/java/org/apache/eventmesh/common/protocol/http/body/client/SubscribeResponseBody.java
by organising the imports.

Documentation
If yes, how is the feature documented? (not applicable / docs / JavaDocs / not documented) Not Applicable
If a feature is not applicable for documentation, explain why? It is a minor issue which comes under code cleanup.


rinuaby13:
Ok will do that

rinuaby13:
Associated this PR to #1741

qqeasonchen (approved):
@rinuaby13 maybe I don't think this is the root cause, but let me merge it first.

PR #1802, opened 2022-10-24
[ISSUE #1692] Method may fail to close stream

### LINKED ISSUE #1692: [Enhancement] Method may fail to close stream
### Search before asking

### Enhancement Request

image
located at:
org/apache/eventmesh/runtime/boot/SSLContextFactory.java
analysis and explanation:
The method creates an IO stream object, does not assign it to any fields, pass it to other methods that might close it, or return it, and does not appear to close the stream on all paths out of the method.  This may result in a file descriptor leak.

Use finally to close resources or try-with-resources to manage resources.

shinnlok:
Fixes #1692.

### Motivation

The method creates an IO stream object, does not assign it to any fields, pass it to other methods that might close it, or return it, and does not appear to close the stream on all paths out of the method. This may result in a file descriptor leak.

### Modifications

Use finally to close resources.


qqeasonchen (approved):
LGTM

PR #1830, opened 2022-10-25
[ISSUE #1699] Refactor. Add NPE check

### LINKED ISSUE #1699: [Enhancement] Possible null pointer dereference due to return value of called method  [EventMeshConsumer]
### Search before asking

### Enhancement Request

image
located at:
org/apache/eventmesh/runtime/core/protocol/http/consumer/EventMeshConsumer.java line 197 198
analysis and explanation:

add NPE exception check.

rinuaby13:
Fixes #1699.

### Motivation
The return value from a method is dereferenced without a null check, and the return value of that method is one that should generally be checked for null. This may lead to a NullPointerException when the code is executed

### Modifications
Added NPE check

### Documentation
- Does this pull request introduce a new feature? (yes / no)No
- If yes, how is the feature documented? (not applicable / docs / JavaDocs / not documented)not applicable
- If a feature is not applicable for documentation, explain why?This is a minor issue that comes under code cleanup


qqeasonchen (approved):
LGTM

Issue #1705, opened 2022-10-21
[Enhancement] Possible null pointer dereference due to return value of called method [SendAsyncMessageProcessor]


Alonexc:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/eventmesh/issues?q=is%3Aissue) and found no similar issues.

### Enhancement Request

![image](https://user-images.githubusercontent.com/91315508/197130621-10fcb234-3c17-4725-8a77-a88d8689b833.png)
![image](https://user-images.githubusercontent.com/91315508/197130634-025f9e29-bba8-4719-b43b-2397647c905e.png)
![image](https://user-images.githubusercontent.com/91315508/197130646-5a9ee8e6-ed47-4cfb-ab05-62083e21cbeb.png)
located at:
org/apache/eventmesh/runtime/core/protocol/http/processor/SendAsyncMessageProcessor.java line 173-175 229 230 234
analysis and explanation:
The return value from a method is dereferenced without a null check, and the return value of that method is one that should generally be checked for null. This may lead to a NullPointerException when the code is executed.

### Describe the solution you'd like

add NPE exception check.

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

NewtonVan:
@xwm1992 Hi, I'm a newbie on eventmesh and I think I'm capable of fixing this issue. Would u pls assign this to me?

PR #1831, opened 2022-10-25
ISSUE (#1705): add NPE exception check

NewtonVan:
- Use `Objects.requireNonNull` achieve fail fast

Fixes [#1705](https://github.com/apache/incubator-eventmesh/issues/1705).

### Motivation

### Modifications
* Added NPE check with `Objects.requireNonNull`
* As for the potential null ptr on line `234`, the data of the event could be empty, so we will return empty string instead of throw exception.

### Documentation

- Does this pull request introduce a new feature? (yes / no) No
- If yes, how is the feature documented? (not applicable / docs / JavaDocs / not documented) Not applicable 
- If a feature is not applicable for documentation, explain why? This is a minor issue that comes under code cleanup


qqeasonchen (approved):
LGTM

Issue #1725, opened 2022-10-21
[Enhancement] Inefficient use of keyset iterator instead of entrySet iterator [LocalUnSubscribeEventProcessor]


Alonexc:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/eventmesh/issues?q=is%3Aissue) and found no similar issues.

### Enhancement Request

![image](https://user-images.githubusercontent.com/91315508/197138650-dca75a2a-d494-4b0f-bd8b-f5702c7ac3eb.png)
located at:
org/apache/eventmesh/runtime/core/protocol/http/processor/LocalUnSubscribeEventProcessor.java 187
analysis and explanation:
This method accesses the value of a Map entry, using a key that was retrieved from a keySet iterator. It is more efficient to use an iterator on the entrySet of the map, to avoid the Map.get(key) lookup.

### Describe the solution you'd like

reference:
![image](https://user-images.githubusercontent.com/91315508/197138805-bcbccd0e-9aa8-4365-b877-e97a46635df8.png)

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

DeepikaMarthandanTW:
I would like to work on this issue. @xwm1992  Can you please assign this task to me?

xwm1992:
> I would like to work on this issue. @xwm1992 Can you please assign this task to me?

Sorry, these issues are labeled as good first issue for first time contributors.

jayabha-jayas:
Hi, I would like to work on this issue.

PR #1913, opened 2022-10-26
[ISSUE #1725] use entrySet replace key set iterator

jayabha-jayas:
Fixes https://github.com/apache/incubator-eventmesh/issues/1725.

Motivation
This method accesses the value of a Map entry, using a key that was retrieved from a keySet iterator. It is more efficient to use an iterator on the entrySet of the map, to avoid the Map.get(key) lookup.

Modifications
we replace key set iteration by using entrySet.

Documentation
If yes, how is the feature documented? (not applicable / docs / JavaDocs / not documented) Not applicable
If a feature is not applicable for documentation, explain why? This is a minor issue that comes under code cleanup.


xwm1992 (approved):
LGTM

Issue #1700, opened 2022-10-21
[Enhancement] Possible null pointer dereference due to return value of called method  [BatchSendMessageProcessor]


Alonexc:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/eventmesh/issues?q=is%3Aissue) and found no similar issues.

### Enhancement Request

located at:
org/apache/eventmesh/runtime/core/protocol/http/processor/BatchSendMessageProcessor.java line 142
analysis and explanation:
The return value from a method is dereferenced without a null check, and the return value of that method is one that should generally be checked for null. This may lead to a NullPointerException when the code is executed.

### Describe the solution you'd like

add NPE exception check.

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

DeepikaMarthandanTW:
I can work on this issue. @xwm1992 could you please assign this to me?

parvathy00:
Hi @Alonexc, I would like to work on this issue. Can you please assign this task to me?

jayabha-jayas:
Hi, I would like to work on this issue

Yamilini2000:
I would like to contribute. Can you please assign me?

Namrata1-99:
I would like to contribute to this issue, can you assign me?

PR #1965, opened 2022-10-27
[ISSUE #1700] add NPE check

jayabha-jayas:
Fixes https://github.com/apache/incubator-eventmesh/issues/1700 .

Motivation

Modifications
If we get null, we will simply return empty string.

Documentation
If yes, how is the feature documented? (not applicable / docs / JavaDocs / not documented) Not applicable
If a feature is not applicable for documentation, explain why? This is a minor issue that comes under code cleanup


xwm1992 (approved):
LGTM

PR #1978, opened 2022-10-27
[ISSUE #1708] add NPE exception check

### LINKED ISSUE #1708: [Enhancement] Possible null pointer dereference due to return value of called method  [ClientSessionGroupMapping]
### Search before asking

### Enhancement Request

image
image
located at:
org/apache/eventmesh/runtime/core/protocol/tcp/client/group/ClientSessionGroupMapping.java line 250 269 275 286-288 300 309 301 329 330 335-337 340-342 347 348 351 352 358 359
analysis and explanation:

add NPE exception check.

NewtonVan:
Fixes #1708 .

### Motivation

### Modifications
- Most of refactoring adapt "fail fast" style.
- Exception check on line 252 is quite different, null cgw won't cause NPE, and it still throw the exception for making `flag` var be false.

### Documentation

* Does this pull request introduce a new feature? (yes / no) no
* If yes, how is the feature documented? (not applicable / docs / JavaDocs / not documented) not applicable
* If a feature is not applicable for documentation, explain why? This is a minor issue that comes under code cleanup


xwm1992 (approved):
LGTM

PR #2285, opened 2022-11-28
[ISSUE #2284]Simplify the code

weihubeats:
Fixes https://github.com/apache/incubator-eventmesh/issues/2284

### Motivation

*Explain the content here.*

### Modifications

### Documentation

### DIFF
--- a/eventmesh-common/src/test/java/org/apache/eventmesh/common/utils/AssertTest.java
+++ b/eventmesh-common/src/test/java/org/apache/eventmesh/common/utils/AssertTest.java
@@ -1,0 +1,19 @@
+package org.apache.eventmesh.common.utils;
+
+import org.junit.Test;
+
+/**
+ * test {@link Assert}
+ */
+public class AssertTest {
+
+    @Test(expected = IllegalArgumentException.class)
+    public void testNotNull() {
+        Assert.notNull(null, "error message");
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void testIsTrue() {
+        Assert.isTrue(false, "error message");
+    }
+}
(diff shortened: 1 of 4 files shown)


xwm1992:
@weihubeats why you add the `Assert` class, junit has already done it.

weihubeats:
Assert is only used for testing, the added Assert is used for null checking in the program, or we can rename the Assert to AssertUtils

xwm1992:
> Assert is only used for testing, the added Assert is used for null checking in the program, or we can rename the Assert to AssertUtils

Ok, maybe rename to `AssertUtils` is better.

weihubeats:
ok

eventmesh-common/src/test/java/org/apache/eventmesh/common/utils/AssertTest.java
```diff
@@ -20,17 +20,17 @@
 import org.junit.Test;
 
 /**
- * test {@link Assert}
+ * test {@link AssertUtils}
  */
 public class AssertTest {
```
jonyangx, line 25: suggest rename AssertUtilsTest.

eventmesh-common/src/test/java/org/apache/eventmesh/common/utils/AssertTest.java
```diff
@@ -20,17 +20,17 @@
 import org.junit.Test;
 
 /**
- * test {@link Assert}
+ * test {@link AssertUtils}
```
jonyangx, line 23: suggest rename AssertUtilsTest.

xwm1992 (approved):

PR #2431, opened 2022-12-03
[ISSUE #2385] return direct result rather than local variable

### LINKED ISSUE #2385: [Enhancement] Method stores return result in local before immediately returning it [HttpSummaryMetrics]
### Search before asking

### Enhancement Request

image
located at:
eventmesh-metrics-plugin/eventmesh-metrics-api/src/main/java/org/apache/eventmesh/metrics/api/model/HttpSummaryMetrics.java line 64 69 95 100 130 164 169 229 234 326 331 390 400 410
analysis and explanation:
This method stores the return result in a local variable, and then immediately returns the local variable. It would be simpler just to return the value that is assigned to the local variable,

reference:
image

keranbingaa:
fix #2385 ,remove local variable before return.


horoc (approved):
LGTM

PR #2441, opened 2022-12-03
[ISSUE #2345]print the log of exception

### LINKED ISSUE #2345: [Enhancement] Method throws alternative exception from catch block without history [ConfigurationWrapper]
### Search before asking

### Enhancement Request

image
located at:
eventmesh-connector-plugin/eventmesh-connector-rabbitmq/src/main/java/org/apache/eventmesh/connector/rabbitmq/config/ConfigurationWrapper.java line 53 61
analysis and explanation:
This method catches an exception, and throws a different exception, without incorporating the original exception. Doing so hides the original source of the exception, making debugging and fixing these problems difficult.

reference:
image
image

keranbingaa:
fix #2345

### DIFF
--- a/eventmesh-connector-plugin/eventmesh-connector-rabbitmq/src/main/java/org/apache/eventmesh/connector/rabbitmq/config/ConfigurationWrapper.java
+++ b/eventmesh-connector-plugin/eventmesh-connector-rabbitmq/src/main/java/org/apache/eventmesh/connector/rabbitmq/config/ConfigurationWrapper.java
@@ -17,8 +17,6 @@
 
 package org.apache.eventmesh.connector.rabbitmq.config;
 
-import lombok.extern.slf4j.Slf4j;
-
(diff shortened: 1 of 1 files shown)


eventmesh-connector-plugin/eventmesh-connector-rabbitmq/src/main/java/org/apache/eventmesh/connector/rabbitmq/config/ConfigurationWrapper.java
```diff
@@ -50,6 +53,7 @@ private static void loadProperties() {
                 PROPERTIES.load(resourceAsStream);
             }
         } catch (IOException e) {
+            log.error("load file from classpath exception:", e);
```
jonyangx, line 56: It's too many same log , I think caller decide log action is better.

xwm1992 (approved):

PR #2509, opened 2022-12-08
[Enhancement] Remove redundant parameter passing

weihubeats:
Fixes https://github.com/apache/incubator-eventmesh/issues/2508

### Motivation

*Explain the content here.*

### Modifications

### Documentation

### DIFF
--- a/eventmesh-common/src/main/java/org/apache/eventmesh/common/config/CommonConfiguration.java
+++ b/eventmesh-common/src/main/java/org/apache/eventmesh/common/config/CommonConfiguration.java
@@ -69,7 +69,7 @@ public class CommonConfiguration {
         if (configurationWrapper != null) {
             eventMeshEnv = checkNotEmpty(ConfKeys.KEYS_EVENTMESH_ENV);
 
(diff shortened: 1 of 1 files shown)


eventmesh-common/src/main/java/org/apache/eventmesh/common/config/CommonConfiguration.java
```diff
@@ -123,12 +123,12 @@ private String checkNotEmpty(String key) {
         return value;
     }
 
-    private String checkNumeric(String key) {
-        String value = configurationWrapper.getProp(key);
+    private String checkNumeric() {
+        String value = configurationWrapper.getProp(ConfKeys.KEYS_EVENTMESH_SYSID);
```
horoc, line 127: I think the method `private String checkNumeric(String key)`  here is introduced as a common method to check the numeric property,  if we remove the key parameter and just use `checkNumeric()` to get specific property,  the method name would be very confusing.

weihubeats: Makes sense

PR #2684, opened 2022-12-24
[ISSUE #2661] Method uses immediate execution of a block of code that is often not used [AbstractEventProcessor]

### LINKED ISSUE #2661: [Enhancement] Method uses immediate execution of a block of code that is often not used [AbstractEventProcessor]
### Search before asking

### Enhancement Request

image
located at:
eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/core/protocol/http/processor/inf/AbstractEventProcessor.java line 129
analysis and explanation:
This method uses the Optional.orElse() method passing in some code that will execute immediately, whether or not the else case of the Optional is needed. This may cause incorrect side effects to happen, or at the minimum, code to execute for no reason.

It would be better to use Optional.orElseGet()

jonyangx:
Fixes #2661 .

### Motivation

Method uses immediate execution of a block of code that is often not used [AbstractEventProcessor]

### Modifications

refactor eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/core/protocol/http/processor/inf/AbstractEventProcessor.java
use Optional.orElseGet()

### Documentation

- If yes, how is the feature documented? ( not documented)

### DIFF
--- a/eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/core/protocol/http/processor/inf/AbstractEventProcessor.java
+++ b/eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/core/protocol/http/processor/inf/AbstractEventProcessor.java
@@ -135,9 +135,7 @@ public class AbstractEventProcessor {
             Map<String, ConsumerGroupTopicMetadata> consumerGroupTopicMetadataMap =
                     Optional.ofNullable(consumerGroupMetadata)
                             .map(ConsumerGroupMetadata::getConsumerGroupTopicMetadataMap)
-                            .orElseGet(() -> {
-                                return Maps.newConcurrentMap();
-                            });
+                            .orElseGet(Maps::newConcurrentMap);
 
             for (SubscriptionItem subscriptionItem : subscriptionList) {
                 if (consumerGroupTopicMetadataMap.containsKey(subscriptionItem.getTopic())) {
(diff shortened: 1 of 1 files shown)


eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/core/protocol/http/processor/inf/AbstractEventProcessor.java
```diff
@@ -125,8 +134,10 @@ protected String getTargetMesh(String consumerGroup, List<SubscriptionItem> subs
                     JsonUtils.deserialize(topicMetadataJson, ConsumerGroupMetadata.class);
             Map<String, ConsumerGroupTopicMetadata> consumerGroupTopicMetadataMap =
                     Optional.ofNullable(consumerGroupMetadata)
-                        .map(ConsumerGroupMetadata::getConsumerGroupTopicMetadataMap)
-                        .orElse(Maps.newConcurrentMap());
+                            .map(ConsumerGroupMetadata::getConsumerGroupTopicMetadataMap)
+                            .orElseGet(() -> {
+                                return Maps.newConcurrentMap();
```
Alonexc, line 139: I think it can be replaced here with .orElseGet(Maps::newConcurrentMap).

jonyangx: Thanks.I've fixed it.

Alonexc (approved):

PR #2793, opened 2023-01-02
[ISSUE #2732] Method calls keySet() just to call contains, use containsKey instead [EventMeshRecommendImpl]

### LINKED ISSUE #2732: [Enhancement] Method calls keySet() just to call contains, use containsKey instead [EventMeshRecommendImpl]
### Search before asking

### Enhancement Request

image
located at:
eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/core/protocol/tcp/client/recommend/EventMeshRecommendImpl.java line 186,192
analysis and explanation:
This method calls mySet.keySet().contains("foo") when mySet.containsKey("foo") is simpler.

reference:
image

jonyangx:
Fixes #2732 .

### Motivation

Method calls keySet() just to call contains, use containsKey instead

### Modifications

replace using Map.keySet().contains with  Map.containsKey().

### Documentation

- If yes, how is the feature documented? (not documented)

### DIFF
--- a/eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/core/protocol/tcp/client/recommend/EventMeshRecommendImpl.java
+++ b/eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/core/protocol/tcp/client/recommend/EventMeshRecommendImpl.java
@@ -36,41 +36,51 @@ import org.slf4j.LoggerFactory;
 
 public class EventMeshRecommendImpl implements EventMeshRecommendStrategy {
 
-    protected final Logger logger = LoggerFactory.getLogger(EventMeshRecommendImpl.class);
+    private static final Logger LOGGER = LoggerFactory.getLogger(EventMeshRecommendImpl.class);
 
-    private EventMeshTCPServer eventMeshTCPServer;
+    private static final int DEFAULT_PROXY_NUM = 1;
 
-    public EventMeshRecommendImpl(EventMeshTCPServer eventMeshTCPServer) {
+    private final transient EventMeshTCPServer eventMeshTCPServer;
+
+    public EventMeshRecommendImpl(final EventMeshTCPServer eventMeshTCPServer) {
         this.eventMeshTCPServer = eventMeshTCPServer;
     }
 
     @Override
-    public String calculateRecommendEventMesh(String group, String purpose) throws Exception {
-        List<EventMeshDataInfo> eventMeshDataInfoList = null;
+    public String calculateRecommendEventMesh(final String group, final String purpose) throws Exception {
+        List<EventMeshDataInfo> eventMeshDataInfoList;
+
         if (StringUtils.isBlank(group) || StringUtils.isBlank(purpose)) {
-            logger.warn("EventMeshRecommend failed,params illegal,group:{},purpose:{}", group, purpose);
+            if (LOGGER.isWarnEnabled()) {
+                LOGGER.warn("EventMeshRecommend failed,params illegal,group:{},purpose:{}", group, purpose);
+            }
             return null;
         }
+
         final String cluster = eventMeshTCPServer.getEventMeshTCPConfiguration().getEventMeshCluster();
         try {
             eventMeshDataInfoList = eventMeshTCPServer.getRegistry().findEventMeshInfoByCluster(cluster);
         } catch (Exception e) {
-            logger.warn("EventMeshRecommend failed, findEventMeshInfoByCluster failed, cluster:{}, group:{}, purpose:{}, errMsg:{}",
-                    cluster, group, purpose, e);
+            if (LOGGER.isWarnEnabled()) {
+                LOGGER.warn("EventMeshRecommend failed, findEventMeshInfoByCluster failed, cluster:{}, group:{}, purpose:{}, errMsg:{}",
+                        cluster, group, purpose, e);
+            }
             return null;
         }
 
         if (eventMeshDataInfoList == null || CollectionUtils.isEmpty(eventMeshDataInfoList)) {
-            logger.warn("EventMeshRecommend failed,not find eventMesh instances from registry,cluster:{},group:{},purpose:{}",
-                    cluster, group, purpose);
+            if (LOGGER.isWarnEnabled()) {
+                LOGGER.warn("EventMeshRecommend failed,not find eventMesh instances from registry,cluster:{},group:{},purpose:{}",
+                        cluster, group, purpose);
+            }
             return null;
         }
 
-        Map<String, String> localEventMeshMap = new HashMap<>();
-        Map<String, String> remoteEventMeshMap = new HashMap<>();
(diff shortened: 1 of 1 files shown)


eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/core/protocol/tcp/client/recommend/EventMeshRecommendImpl.java
```diff
@@ -31,179 +31,233 @@
 import java.util.List;
 import java.util.Map;
 
-import lombok.extern.slf4j.Slf4j;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
-@Slf4j
```
horoc, line 36: hi, @jonyangx , `slf4j` log annotation here was introduced in this PR #2761 recently by @weihubeats , may I know why do you want to remove it? 
maybe we need to discuss and make a consensus to use a unified solution for the whole project.

jonyangx: lombok is not good for ide. @horoc

weihubeats: I think since we use lombok we can use @ slf4j to simplify the log printing

horoc: Actually, we do have lots of Lombok annotations in the project, like `@Data`, `@Builder`, `@Slf4j`,  and as @jonyangx mentioned, it's true that there is some shortage of Lombok annotation, the developers need to do extra configuration in IDE to enable Lombok parsing. But in the meantime, Lombok also has provided its own convenience for developers.

I think the key point is, we need to keep a consistent code style for the whole project, otherwise,  we may modify the same place back and forth by different developers. 😂

jonyangx: OK, I could fix it using lombok,but it's not good for vim/ultraedit ide.

jonyangx: @weihubeats  how to simplify the log printing?

jonyangx: I don't think so.

horoc: > OK, I could fix it using lombok,but it's not good for vim/ultraedit ide.

@jonyangx , I think your concern about the inconvenience of developers who use other IDE is reasonable.

In my point of view, if you think it's better to remove Lombok dependencies,  you can raise a new issue to discuss this point, and after discussion, if it's really necessary, we can do an optimization in a new pr to remove all Lombok annotations.

weihubeats: If we want to print the log we just need to add the comment @Slf4j, without writing similar code
`
private static final Logger LOG = LoggerFactory.getLogger(WeightRoundRobinLoadBalanceSelector.class);
`
In practice, sometimes we will find some complex code xx.class in the project is wrong, and not the actual printing log class, so that we can reduce the workload of printing logs and improve development efficiency

weihubeats: If it is not friendly to the development tools vim/ultraedit, how should the heavy use of lombok in the project be handled? Should the entire code style be consistent, either by removing lombok dependencies or by using lombok uniformly?

eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/core/protocol/tcp/client/recommend/EventMeshRecommendImpl.java
```diff
@@ -31,63 +31,82 @@
...
+                log.warn("EventMeshRecommend failed, findEventMeshInfoByCluster failed, cluster:{}, group:{}, purpose:{}, errMsg:{}",
+                        cluster, group, purpose, e);
+            }
             return null;
         }
 
-        if (CollectionUtils.isEmpty(eventMeshDataInfoList)) {
-            log.warn("EventMeshRecommend failed,not find eventMesh instances from registry,cluster:{},group:{},purpose:{}",
-                    cluster, group, purpose);
+        if (eventMeshDataInfoList == null || CollectionUtils.isEmpty(eventMeshDataInfoList)) {
```
horoc, line 72: image

`CollectionUtils.isEmpty` has involved the null pointer check

eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/core/protocol/tcp/client/recommend/EventMeshRecommendImpl.java
```diff
@@ -31,63 +31,82 @@
...
     }
 
     @Override
-    public String calculateRecommendEventMesh(String group, String purpose) {
+    public String calculateRecommendEventMesh(final String group, final String purpose) throws Exception {
         List<EventMeshDataInfo> eventMeshDataInfoList;
-        if (StringUtils.isAnyBlank(group, purpose)) {
-            log.warn("EventMeshRecommend failed,params illegal,group:{},purpose:{}", group, purpose);
+
+        if (StringUtils.isBlank(group) || StringUtils.isBlank(purpose)) {
```
horoc, line 54: May I know what's the difference between using `isAnyBlank` and `isBlank` here?  I think the functionality is the same.

eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/core/protocol/tcp/client/recommend/EventMeshRecommendImpl.java
jonyangx: I have fixed it. @horoc

eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/core/protocol/tcp/client/recommend/EventMeshRecommendImpl.java
jonyangx: I have fixed it. @horoc

horoc (approved):

PR #2800, opened 2023-01-03
[ISSUE #2728] Method check a map with containsKey(), before using get() [SubscribeProcessor]

### LINKED ISSUE #2728: [Enhancement] Method check a map with containsKey(), before using get() [SubscribeProcessor]
### Search before asking

### Enhancement Request

image
located at:
eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/core/protocol/http/processor/SubscribeProcessor.java line 206
analysis and explanation:
This method checks for the presence of a key in a map using containsKey(), before attempting to fetch the value of the key using get(). This equates to doing two map lookups in a row.

It is much simpler to just fetch the value with get, and checking for non null instead.

mvadharia:
Fixes #2728 .

### Motivation

*This method checks for the presence of a key in a map using containsKey(), before attempting to fetch the value of the key using get(). This equates to doing two map lookups in a row.*

### Modifications

*Refactored code to do mapLookUp once and then check non null , if null then create a new list else add to existing list.*

### Documentation


qqeasonchen (approved):
LGTM

PR #2848, opened 2023-01-06
[ISSUE #2822] Use retainAll method instead of iterating over keys in Map [EventMeshRecommendImpl]

### LINKED ISSUE #2822: [Enhancement] Use retainAll method instead of iterating over keys in Map [EventMeshRecommendImpl]
### Search before asking

### Enhancement Request

located at:
eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/core/protocol/tcp/client/recommend/EventMeshRecommendImpl.java line 176

Iteration over keyset can be avoided.

Using the retainAll method to get the intersection of  sets from the keySet method of the Map interface to get the set of keys for both map1 and map2 can avoid iteration over entire keySet.

mvadharia:
Fixes #2822 .

### Motivation

*Iteration over keyset can be avoided or use of containsAll could be avoided*

### Modifications

*Using the retainAll method to get the intersection of sets from the keySet method of the Map interface to get the set of keys for both map1 and map2 can avoid iteration over entire keySet.*

### Documentation

- Does this pull request introduce a new feature? ( no)

### DIFF
--- a/eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/core/protocol/tcp/client/recommend/EventMeshRecommendImpl.java
+++ b/eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/core/protocol/tcp/client/recommend/EventMeshRecommendImpl.java
@@ -230,7 +230,7 @@ public class EventMeshRecommendImpl implements EventMeshRecommendStrategy {
 
         Set<String> clientdDistributionSet=new HashSet<>(clientDistributionMap.keySet());
         clientdDistributionSet.retainAll(eventMeshMap.keySet());
-        if(!clientdDistributionSet.isEmpty()){
+        if(clientdDistributionSet.isEmpty()){
             if (log.isWarnEnabled()) {
                 log.warn("exist proxy not register but exist in distributionMap");
             }
--- a/eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/core/protocol/tcp/client/recommend/EventMeshRecommendImpl.java
+++ b/eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/core/protocol/tcp/client/recommend/EventMeshRecommendImpl.java
@@ -25,7 +25,14 @@ import org.apache.commons.collections4.CollectionUtils;
 import org.apache.commons.collections4.MapUtils;
 import org.apache.commons.lang3.StringUtils;
 
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+import java.util.HashSet;
 
 import lombok.extern.slf4j.Slf4j;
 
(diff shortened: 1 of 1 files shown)


eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/core/protocol/tcp/client/recommend/EventMeshRecommendImpl.java
```diff
@@ -226,15 +228,16 @@ private String recommendProxy(final Map<String, String> eventMeshMap,
             log.info("eventMeshMap:{},clientDistributionMap:{},group:{}", eventMeshMap, clientDistributionMap, group);
         }
 
-        if (!eventMeshMap.keySet().containsAll(clientDistributionMap.keySet())) {
+        Set<String> clientdDistributionSet=new HashSet<>(clientDistributionMap.keySet());
+        clientdDistributionSet.retainAll(eventMeshMap.keySet());
+        if(!clientdDistributionSet.isEmpty()){
```
jonyangx, line 233: The retainAll() method of java.util.Set interface is used to retain from this set all of its elements that are contained in the specified collection.
I think the code logic is invalid.

eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/core/protocol/tcp/client/recommend/EventMeshRecommendImpl.java
```diff
@@ -226,15 +228,16 @@ private String recommendProxy(final Map<String, String> eventMeshMap,
             log.info("eventMeshMap:{},clientDistributionMap:{},group:{}", eventMeshMap, clientDistributionMap, group);
         }
 
-        if (!eventMeshMap.keySet().containsAll(clientDistributionMap.keySet())) {
+        Set<String> clientdDistributionSet=new HashSet<>(clientDistributionMap.keySet());
+        clientdDistributionSet.retainAll(eventMeshMap.keySet());
```
jonyangx, line 232: Retains only the elements in this collection that are contained in the specified collection (optional operation). In other words, removes from this collection all of its elements that are not contained in the specified collection.
This implementation iterates over this collection, checking each element returned by the iterator in turn to see if it's contained in the specified collection. If it's not so contained, it's removed from this collection with the iterator's remove method.
I think Iteration over keyset can not be avoided .

mvadharia: @jonyangx  Idea was to use retainAll() method to get intersection of two sets of keys, if there exists one element in intersection set which indirectly proves there exist proxy(key) not register(i.e. not present in eventMeshMap) but exist in distributionMap"

eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/core/protocol/tcp/client/recommend/EventMeshRecommendImpl.java
mvadharia: Need to invert it to get correct logic when using retainAll()

eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/core/protocol/tcp/client/recommend/EventMeshRecommendImpl.java
jonyangx: public boolean removeAll(Collection<?> c) {
        Objects.requireNonNull(c);
        boolean modified = false;
        Iterator<?> it = iterator();
        while (it.hasNext()) {
            if (c.contains(it.next())) {
                it.remove();
                modified = true;
            }
        }
        return modified;
    }

The removeAll method also uses iterator()  , and use remove operation.  Why not use containAll directly?

mvadharia: Ok. Got it. Then would close the PR and issue.

PR #2924, opened 2023-01-13
[ISSUE #1538] Fixes issue with integer-based for loop in BatchMessage.java

### LINKED ISSUE #1538: [Enhancement] Method uses integer based for loops to iterate over a List [BatchMessage]
### Search before asking

### Enhancement Request

image
located at: org/apache/eventmesh/common/protocol/grpc/protos/BatchMessage.java line 1668 and 1690
explanation:This method uses an integer-based for loop to iterate over a java.util.List, by calling List.get(i) each time through the loop. The integer is not used for other reasons. It is better to use an Iterator instead, as depending on List implementation, iterators can perform better, and they also allow for exchanging of other collection types without issue.

It is recommended to use iterators or foreach.

Orcohen33:
Fixes #1538 .

### Motivation

The method uses an integer-based for loop to iterate over a java.util.List, by calling List.get(i) each time through the loop. This can lead to poor performance and issues when using other collection types. The goal of this change is to improve performance and flexibility by using iterators or foreach loops instead.

### Modifications

Replaced the integer-based for loop with an iterator or foreach loop in the method located at org/apache/eventmesh/common/protocol/grpc/protos/BatchMessage.java line 1558 and 1560

lines1558-1560

### Documentation

- If a feature is not applicable for documentation, explain why? This pull request does not introduce a new feature, it only improves the performance and flexibility of the existing code.
- If a feature is not documented yet in this PR, please create a followup issue for adding the documentation (not applicable)

### DIFF
--- a/eventmesh-common/src/main/java/org/apache/eventmesh/common/protocol/grpc/protos/BatchMessage.java
+++ b/eventmesh-common/src/main/java/org/apache/eventmesh/common/protocol/grpc/protos/BatchMessage.java
@@ -1555,9 +1555,7 @@ public final class BatchMessage extends
                 (!getTopicBytes().isEmpty() ?
                         com.google.protobuf.GeneratedMessageV3.computeStringSize(3, topic_) : 0);
 
-        for (MessageItem messageItem : messageItem_) {
-            size += com.google.protobuf.CodedOutputStream.computeMessageSize(4, messageItem);
-        }
+        messageItem_.forEach(messageItem -> size += com.google.protobuf.CodedOutputStream.computeMessageSize(4, messageItem));
 
         size += unknownFields.getSerializedSize();
         memoizedSize = size;
--- a/eventmesh-common/src/main/java/org/apache/eventmesh/common/protocol/grpc/protos/BatchMessage.java
+++ b/eventmesh-common/src/main/java/org/apache/eventmesh/common/protocol/grpc/protos/BatchMessage.java
@@ -1555,9 +1555,8 @@ public final class BatchMessage extends
                 (!getTopicBytes().isEmpty() ?
                         com.google.protobuf.GeneratedMessageV3.computeStringSize(3, topic_) : 0);
 
-        for (int i = 0; i < messageItem_.size(); i++) {
-            size += com.google.protobuf.CodedOutputStream
-                    .computeMessageSize(4, messageItem_.get(i));
+        for (MessageItem messageItem : messageItem_) {
+            size += com.google.protobuf.CodedOutputStream.computeMessageSize(4, messageItem);
         }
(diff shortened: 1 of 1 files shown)


eventmesh-common/src/main/java/org/apache/eventmesh/common/protocol/grpc/protos/BatchMessage.java
```diff
@@ -1555,9 +1555,8 @@ public int getSerializedSize() {
                 (!getTopicBytes().isEmpty() ?
                         com.google.protobuf.GeneratedMessageV3.computeStringSize(3, topic_) : 0);
 
-        for (int i = 0; i < messageItem_.size(); i++) {
-            size += com.google.protobuf.CodedOutputStream
-                    .computeMessageSize(4, messageItem_.get(i));
+        for (MessageItem messageItem : messageItem_) {
```
jonyangx, line 1558: @Orcohen33 suggest use reactor pattern ,use messageItem_.foreach().

Orcohen33: Hey @jonyangx , I've changed it again and pushed it.

jonyangx:
@Orcohen33 pls fix build error.

Orcohen33:
> @Orcohen33 pls fix build error.

@jonyangx I'm sorry for the inconvenience, I've found an issue in the code that's causing a build error. 

The problem is located at the following file: org/apache/eventmesh/common/protocol/grpc/protos/BatchMessage.java line 1558 and 1560

I've implemented the forEach method as you suggested. However, I encountered an issue where I'm not able to modify a variable defined outside the forEach block.

To work around this limitation, I've used an array or an ArrayList to store the computed sizes and then used the reduce method to sum all the sizes. Here's an example of the implementation:

```java
List<Integer> sizes = new ArrayList<>();
messageItem_.forEach(item -> {
    sizes.add(com.google.protobuf.CodedOutputStream.computeMessageSize(4, item));
});
size = sizes.stream().reduce(0, (a, b) -> a + b);
```
or
```java
AtomicInteger size = new AtomicInteger();
messageItem_.forEach(item -> {
    size.addAndGet(com.google.protobuf.CodedOutputStream.computeMessageSize(4, item));
});
```
This solution worked for me, but please keep in mind that it's important to test the code and make sure it works as expected.

Please let me know if that's okay and I'll submit this change for review.

jonyangx:
@Orcohen33 I think the second segment code is better.

jonyangx:
@Orcohen33 I think the second segment code is better.

jonyangx (approved):

PR #2955, opened 2023-01-24
Issue #2715 Update ClientSessionGroupMapping.java

### LINKED ISSUE #2715: [Enhancement] Method checks the size of a collection against zero rather than using isEmpty() [ClientSessionGroupMapping]
### Search before asking

### Enhancement Request

located at:
eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/core/protocol/tcp/client/group/ClientSessionGroupMapping.java line 331,337,338
analysis and explanation:
This method calls the size() method on a collection and compares the result to zero to see if the collection is empty.

For better code clarity, it is better to just use col.isEmpty().

PreetAngadSingh:
Fixes #2715 

Motivation
Top reasons to use isEmpty instead of size()

it is more expressive (the code is easier to read and to maintain)
it is faster, in some cases by orders of magnitude.
Modifications
Changed the size() method on a collection to isEmpty()

Documentation
Does this pull request introduce a new feature? (no)

### DIFF
--- a/eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/core/protocol/tcp/client/group/ClientSessionGroupMapping.java
+++ b/eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/core/protocol/tcp/client/group/ClientSessionGroupMapping.java
@@ -323,14 +323,14 @@ public class ClientSessionGroupMapping {
     private void cleanClientGroupWrapperCommon(ClientGroupWrapper clientGroupWrapper) throws Exception {
         log.info("GroupConsumerSessions size:{}",
                 clientGroupWrapper.getGroupConsumerSessions().size());
-        if (clientGroupWrapper.getGroupConsumerSessions().size() == 0) {
+        if (clientGroupWrapper.getGroupConsumerSessions().isEmpty()) {
             shutdownClientGroupConsumer(clientGroupWrapper);
         }
 
         log.info("GroupProducerSessions size:{}",
                 clientGroupWrapper.getGroupProducerSessions().size());
-        if ((clientGroupWrapper.getGroupConsumerSessions().size() == 0)
-                && (clientGroupWrapper.getGroupProducerSessions().size() == 0)) {
+        if ((clientGroupWrapper.getGroupConsumerSessions().isEmpty())
+                && (clientGroupWrapper.getGroupProducerSessions().isEmpty())) {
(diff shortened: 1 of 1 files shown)


eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/core/protocol/tcp/client/group/ClientSessionGroupMapping.java
```diff
@@ -323,14 +323,14 @@ private void handleUnackMsgsInSession(Session session) {
     private void cleanClientGroupWrapperCommon(ClientGroupWrapper clientGroupWrapper) throws Exception {
         log.info("GroupConsumerSessions size:{}",
                 clientGroupWrapper.getGroupConsumerSessions().size());
-        if (clientGroupWrapper.getGroupConsumerSessions().size() == 0) {
+        if (clientGroupWrapper.getGroupConsumerSessions().isEmpty()) {
```
jonyangx, line 326: HashMap source code:
  /**
     * Returns the number of key-value mappings in this map.
     *
     * @return the number of key-value mappings in this map
     */
    public int size() {
        return size;
    }

    /**
     * Returns <tt>true</tt> if this map contains no key-value mappings.
     *
     * @return <tt>true</tt> if this map contains no key-value mappings
     */
    public boolean isEmpty() {
        return size == 0;
    }

Suggest check clientGroupWrapper.getGroupConsumerSessions() is Null?

PreetAngadSingh: @jonyangx please review this pull request!

PreetAngadSingh: I am having trouble with this commit... please guide me how to correct this! @jonyangx

jonyangx:
@PreetAngadSingh 
change import order ,
......
import org.apache.eventmesh.runtime.util.RemotingHelper;

import org.apache.commons.collections4.CollectionUtils;
import org.apache.commons.collections4.MapUtils;

import java.lang.ref.WeakReference;
......

jonyangx (approved):

PreetAngadSingh:
Thank you so much @jonyangx

PR #3024, opened 2023-02-02
[ISSUE #3020]Get String bytes by StandardCharsets.UTF_8

### LINKED ISSUE #3020: [Enhancement] Found reliance on default encoding: String.getBytes() [TopicHandler]
### Search before asking

### Enhancement Request

image
located at:
eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/admin/handler/TopicHandler.java line 88,89,99,100,136,137
analysis and explanation:

e.g.
image

mxsm:
Fixes #3020 

### Motivation

*Explain the content here.*

### Modifications

- Get String bytes by StandardCharsets.UTF_8

### Documentation

### DIFF
--- a/eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/admin/handler/TopicHandler.java
+++ b/eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/admin/handler/TopicHandler.java
@@ -18,6 +18,7 @@
 package org.apache.eventmesh.runtime.admin.handler;
 
 import org.apache.eventmesh.api.admin.TopicProperties;
+import org.apache.eventmesh.common.Constants;
 import org.apache.eventmesh.runtime.admin.controller.HttpHandlerManager;
 import org.apache.eventmesh.runtime.admin.request.CreateTopicRequest;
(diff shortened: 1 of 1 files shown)


eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/admin/handler/TopicHandler.java
```diff
@@ -85,8 +86,8 @@ void get(HttpExchange httpExchange) throws IOException {
         try {
             List<TopicProperties> topicList = admin.getTopic();
             String result = JsonUtils.toJson(topicList);
-            httpExchange.sendResponseHeaders(200, result.getBytes().length);
-            out.write(result.getBytes());
+            httpExchange.sendResponseHeaders(200, result.getBytes(StandardCharsets.UTF_8).length);
```
mytang0, line 89: I remember that the public module should have the definition of charsets, which you can use directly.

mxsm:
@mytang0  PTAL~

xwm1992 (approved):

Issue #3181, opened 2023-02-17
[Enhancement] The method parameter properties of the aclService is adjusted to an object parameter, providing extended fields


slowpao:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/eventmesh/issues?q=is%3Aissue) and found no similar issues.

### Enhancement Request

The method parameter properties of aclService cannot support object passing other than strings, so it is adjusted to provide extended fields for object parameters

### Describe the solution you'd like

The method parameter properties of aclService cannot support object passing other than strings, so it is adjusted to provide extended fields for object parameters

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #3185, opened 2023-02-17
Fix 3181

slowpao:
Fix 3181

Modifications
Encapsulates the receive parameters of methods in the aclService Service service as objects and provides extended fields


mxsm (commented):
Please resolve the conflict @slowpao  and link this pr to issue like this: Fix #3181 in first comment

PR #3206, opened 2023-02-19
[ISSUE #3168] IO streams closed, encoding set to UTF_8, test fails if file change not occurred

### LINKED ISSUE #3168: [Enhancement] Method may fail to close stream[WatchFileManagerTest]
### Search before asking

### Enhancement Request

image
located at:
eventmesh-common/src/test/java/org/apache/eventmesh/common/file/WatchFileManagerTest.java line 54,56
analysis and explanation:
1.The method creates an IO stream object, does not assign it to any fields, pass it to other methods that might close it, or return it, and does not appear to close the stream on all paths out of the method.  This may result in a file descriptor leak. 
2. Found a call to a method which will perform a byte to String (or String to byte) conversion, and will assume that the default platform encoding is suitable. This will cause the application behaviour to vary between platforms.

1.It's a good idea to use try-with-resource to manage resources.
2. Use an alternative API and specify a charset name or Charset object explicitly.

DimuthuWeerathunga:
Fixes #3168.

### Modifications

I have changed the api s used to read to and write from the file
previous --> Line 54 and 56
updated --> Line 54 and 55

made sure the resources are closed by enclosing it with try with resources.

made sure the test fails if it fails to read from the given file or  fails to write to it.

### Documentation

### DIFF
--- a/eventmesh-common/src/test/java/org/apache/eventmesh/common/file/WatchFileManagerTest.java
+++ b/eventmesh-common/src/test/java/org/apache/eventmesh/common/file/WatchFileManagerTest.java
@@ -22,8 +22,6 @@ import org.apache.eventmesh.common.utils.ThreadUtils;
 
 import java.io.*;
 import java.nio.charset.StandardCharsets;
-import java.nio.file.Path;
-import java.nio.file.Paths;
 import java.util.Properties;
 import java.util.concurrent.TimeUnit;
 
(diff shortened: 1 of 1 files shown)


horoc (commented):
image

please fix the checkstyle error

DimuthuWeerathunga:
#3168 

i can't exactly pinpoint what the bug is but when i include the BufferedWriter and BufferedReader inside the brackets of try clause (try with resources) it fails that given test. when i dont close the reader and writer the tests succeed

DimuthuWeerathunga:
This is the only way i could get this to work.
By changing the api slightly and closing the streams manually in finally blcok

PR #3216, opened 2023-02-20
[ISSUE #3155] Method passes a concatenated string to SLF4J's or Log4j2's format string[CClientDemo] 

### LINKED ISSUE #3155: [Enhancement]  Method passes a concatenated string to SLF4J's or Log4j2's format string[CClientDemo]
### Search before asking

### Enhancement Request

image
located at:
eventmesh-runtime/src/test/java/org/apache/eventmesh/runtime/demo/CClientDemo.java line 57
analysis and explanation:
This method uses an SLF4J or Log4j2 logger to log a string, where the first (format) string is created using concatenation. You should use {} markers to inject dynamic content into the string, so that String building is delayed until the actual log string is needed. If the log level is high enough that this log statement isn't used, then the appends will never be executed.

image

Bannirui:
Fixes #3155 .

### Modifications

Locating at eventmesh-runtime/src/test/java/org/apache/eventmesh/runtime/demo/CClientDemo.java

- Line 54: Replace `==` with `Objects::equals`
- Line 57: Logging with placeholder

### Documentation

### DIFF
--- a/eventmesh-runtime/src/test/java/org/apache/eventmesh/runtime/demo/CClientDemo.java
+++ b/eventmesh-runtime/src/test/java/org/apache/eventmesh/runtime/demo/CClientDemo.java
@@ -25,6 +25,8 @@ import org.apache.eventmesh.runtime.client.common.MessageUtils;
 import org.apache.eventmesh.runtime.client.hook.ReceiveMsgHook;
 import org.apache.eventmesh.runtime.client.impl.EventMeshClientImpl;
 
+import java.util.Objects;
+
 import io.netty.channel.ChannelHandlerContext;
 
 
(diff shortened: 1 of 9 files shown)


eventmesh-runtime/src/test/java/org/apache/eventmesh/runtime/demo/CClientDemo.java
```diff
@@ -49,11 +51,10 @@ public static void main(String[] args) throws Exception {
         client.registerSubBusiHandler(new ReceiveMsgHook() {
             @Override
             public void handle(Package msg, ChannelHandlerContext ctx) {
-                if (msg.getHeader().getCmd() == Command.ASYNC_MESSAGE_TO_CLIENT
-                        || msg.getHeader().getCmd() == Command.BROADCAST_MESSAGE_TO_CLIENT) {
-
+                if (Objects.equals(msg.getHeader().getCmd(), Command.ASYNC_MESSAGE_TO_CLIENT)
```
jonyangx, line 54: Therefore we can conclude that when we use == operator for enum member comparison then it performs type compatibility check at compile time whereas .equals() method will never worry about the types of both the arguments.So suggest use == operator.

xwm1992 (approved):

PR #3218, opened 2023-02-20
[ISSUE #3082] Replaced orElse() with orElseGet() on line 61

### LINKED ISSUE #3082: [Enhancement] Method uses immediate execution of a block of code that is often not used
### Search before asking

### Enhancement Request

image
located at:
eventmesh-common/src/main/java/org/apache/eventmesh/common/config/convert/ConvertValue.java line 61
analysis and explanation:
This method uses the Optional.orElse() method passing in some code that will execute immediately, whether or not the else case of the Optional is needed. This may cause incorrect side effects to happen, or at the minimum, code to execute for no reason.

It would be better to use Optional.orElseGet().
e.g.
image

PickBas:
Fixes #3082.

### Modifications

Replaced orElse() with orElseGet() on line 61

### Documentation


mytang0 (approved):
LGTM

PR #3219, opened 2023-02-20
[ISSUE #3181] add aclProperties Object

### LINKED ISSUE #3181: [Enhancement] The method parameter properties of the aclService is adjusted to an object parameter, providing extended fields

slowpao:
Fixes #3181 
The AclService interface has been modified so that its methods can receive custom objects

### DIFF
--- a/eventmesh-security-plugin/eventmesh-security-api/src/main/java/org/apache/eventmesh/api/acl/AclProperties.java
+++ b/eventmesh-security-plugin/eventmesh-security-api/src/main/java/org/apache/eventmesh/api/acl/AclProperties.java
@@ -1,0 +1,105 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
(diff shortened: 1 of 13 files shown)


eventmesh-security-plugin/eventmesh-security-api/src/main/java/org/apache/eventmesh/api/acl/AclProperties.java
```diff
@@ -0,0 +1,105 @@
...
+
+    private String clientIp;
+    private String user;
+    private String pwd;
+    private String subsystem;
+    private String topic;
+    private Integer requestCode;
+    private String requestURI;
+    private String token;
+    private ConcurrentHashMap<String, Object> extendedField;
```
xwm1992, line 32: Add this map field, but never set/get the map in this pull request, perhaps you need also expand the interface or method.

slowpao: Yes, I would resubmit the extension method for manipulating map

Issue #3255, opened 2023-02-24
[Enhancement] Synchronization on a non-final field 'sender'


Alonexc:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/eventmesh/issues?q=is%3Aissue) and found no similar issues.

### Enhancement Request

![image](https://user-images.githubusercontent.com/91315508/221105069-21c09e34-88d3-4f58-ab5d-e4c5b2cf2df5.png)
located at:
eventmesh-sdk-java/src/main/java/org/apache/eventmesh/client/grpc/consumer/SubStreamHandler.java line 162,172

### Describe the solution you'd like

![image](https://user-images.githubusercontent.com/91315508/221105096-3929277e-e1a7-49d0-b3ee-c3a34ea40036.png)

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

PR #3281, opened 2023-02-27
[#3255] removed Synchronization

nikam14:
made changes mentioned in issue #3255 .
open for help.


eventmesh-sdk-java/src/main/java/org/apache/eventmesh/client/grpc/consumer/SubStreamHandler.java
```diff
@@ -159,19 +159,15 @@ public void close() {
 
     private void senderOnNext(final Subscription subscription) {
```
Alonexc, line 160: My suggestion is to put 'synchronized' on the method. Do you have any better suggestions?

nikam14: Yes I am thinking same of putting back the 'synchronized' on the method

PR #3320, opened 2023-03-01
[ISSUE #3084] Reliance on default encoding[FileLoad]

### LINKED ISSUE #3084: [Enhancement] Reliance on default encoding[FileLoad]
### Search before asking

### Enhancement Request

image
located at:
eventmesh-common/src/main/java/org/apache/eventmesh/common/config/FileLoad.java line 71,73
analysis and explanation:
1.Found a call to a method which will perform a byte to String (or String to byte) conversion, and will assume that the default platform encoding is suitable. This will cause the application behaviour to vary between platforms. Use an alternative API and specify a charset name or Charset object explicitly.
2. The method creates an IO stream object, does not assign it to any fields, pass it to other methods that might close it, or return it, and does not appear to close the stream on all paths out of the method.  This may result in a file descriptor leak.  It is generally a good idea to use a finally block to ensure that streams are closed.

1.Use try-with-resources to manage resources.
2. Use an alternative API and specify a charset name or Charset object explicitly.

sanskar-thakur:
Fixes #3084 .

### Modifications

1. Specified charset with input-stream-reader API.
2. Used try-with-resources and handled io stream close.

### Documentation

- Does this pull request introduce a new feature? **(no)**
- If yes, how is the feature documented? **(not applicable)**
- If a feature is not applicable for documentation, explain why? **(bug fix)**
- If a feature is not documented yet in this PR, please create a followup issue for adding the documentation **(not applicable)**

### DIFF
--- a/eventmesh-common/src/main/java/org/apache/eventmesh/common/config/FileLoad.java
+++ b/eventmesh-common/src/main/java/org/apache/eventmesh/common/config/FileLoad.java
@@ -17,8 +17,6 @@
 
 package org.apache.eventmesh.common.config;
 
-import java.io.FileNotFoundException;
-import lombok.extern.slf4j.Slf4j;
 import org.apache.eventmesh.common.Constants;
 import org.apache.eventmesh.common.config.convert.Convert;
 
@@ -27,9 +25,10 @@ import org.apache.commons.lang3.StringUtils;
 import java.io.BufferedInputStream;
 import java.io.BufferedReader;
 import java.io.FileInputStream;
-import java.io.FileReader;
 import java.io.IOException;
 import java.io.InputStreamReader;
+import java.nio.file.Files;
+import java.nio.file.Paths;
 import java.util.Objects;
 import java.util.Properties;
 
@@ -63,7 +62,6 @@ public interface FileLoad {
 
     <T> T getConfig(ConfigInfo configInfo) throws IOException;
 
-    @Slf4j
     class PropertiesFileLoad implements FileLoad {
 
         private final Convert convert = new Convert();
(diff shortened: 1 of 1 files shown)


xwm1992:
image
@sanskar-thakur please fix the ci check error, take care of package import order.

mxsm (commented):
@sanskar-thakur  When you're done you can check the format using the checkStyle plugin or running Gradle's checkstyleMain or checkstyleTest commands.

eventmesh-common/src/main/java/org/apache/eventmesh/common/config/FileLoad.java
```diff
@@ -68,9 +72,28 @@ class PropertiesFileLoad implements FileLoad {
...
             if (StringUtils.isNotBlank(configInfo.getResourceUrl())) {
-                properties.load(new BufferedReader(new InputStreamReader(getClass().getResourceAsStream(configInfo.getResourceUrl()))));
+                try (InputStreamReader reader = new InputStreamReader(
+                    Objects.requireNonNull(getClass().getResourceAsStream(configInfo.getResourceUrl())), Constants.DEFAULT_CHARSET)) {
+                    properties.load(new BufferedReader(reader));
+                }
             } else {
-                properties.load(new BufferedReader(new FileReader(configInfo.getFilePath())));
+                FileReader reader = null;
+                try {
```
Alonexc, line 81: Here you can also change to use try-with-resources, as well as specify the character set.
I have a suggestion, how about you look at:
image

sanskar-thakur: Agreed. Updated this code block with try-with-resources.

Alonexc (approved):
LGTM

xwm1992 (approved):

PR #3486, opened 2023-03-19
[ISSUE #2974] logging concat update

### LINKED ISSUE #2974: [Enhancement] Method passes a concatenated string to SLF4J's or Log4j2's format string[CloudEventProducer]
### Search before asking

### Enhancement Request

image
located at:
eventmesh-sdk-java/src/main/java/org/apache/eventmesh/client/grpc/producer/CloudEventProducer.java line 75,88,97,110,119
analysis and explanation:
1. This method uses an SLF4J or Log4j2 logger to log a string, where the first (format) string is created using concatenation. You should use {} markers to inject dynamic content into the string, so that String building is delayed until the actual log string is needed. If the log level is high enough that this log statement isn't used, then the appends will never be executed.
2. This method concatenates the output of a toString() call into a StringBuffer or StringBuilder. It is simpler just to pass the object you want to append to the append call, as that form does not suffer the potential for NullPointerExceptions, and is easier to read.

Proper use of SLF4J for logging information, and removal of .toString to avoid NPE.
e.g.
image

kvaithin:
Fixes #2974

### Modifications

Update to remove concatenation and instead use placeholders for logging.

### DIFF
--- a/eventmesh-sdk-java/src/main/java/org/apache/eventmesh/client/grpc/producer/CloudEventProducer.java
+++ b/eventmesh-sdk-java/src/main/java/org/apache/eventmesh/client/grpc/producer/CloudEventProducer.java
@@ -74,7 +74,7 @@ public class CloudEventProducer {
         try {
             final Response response = publisherClient.batchPublish(enhancedMessage);
             if (log.isInfoEnabled()) {
-                log.info("Received response " + response.toString());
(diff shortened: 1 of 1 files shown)


mxsm (approved):
LGTM~

Alonexc (approved):
LGTM

xwm1992 (approved):

PR #3624, opened 2023-04-03
Refactor: No changes to code processing logic

### LINKED ISSUE #3388: [Enhancement] The code processing logic here is the same.
### Search before asking

### Enhancement Request

image
located at:
eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/core/protocol/http/push/AsyncHTTPPushRequest.java line 233-247
analysis and explanation:
If the same logic is truly needed for both instances, then in an if chain they should be combined, or for a switch, one should fall through to the other.

image

kyooosukedn:
Fixes #3388.

### Motivation

*The code processing logic here is the same. If the same logic is truly needed for both instances, then in an if chain they should be combined, or for a switch, one should fall through to the other.*

### Modifications

*Join two if statements that perform same code logic.*

### Documentation


Alonexc:
LGTM

mytang0 (approved):

PR #3630, opened 2023-04-05
[ISSUE #3483]Replace this call to "replaceAll()" by a call to the "replace()" method.[PushConsumerImpl]

### LINKED ISSUE #3483: [Enhancement] Replace this call to "replaceAll()" by a call to the "replace()" method.[PushConsumerImpl]
### Search before asking

### Enhancement Request

image
located at:
eventmesh-storage/eventmesh-storage-rocketmq/src/main/java/org/apache/eventmesh/storage/rocketmq/consumer/PushConsumerImpl.java line 189,260
analysis and explanation:
The underlying implementation of String::replaceAll calls the java.util.regex.Pattern.compile() method each time it is called even if the first argument is not a regular expression. This has a significant performance cost and therefore should be used with care.
When String::replaceAll is used, the first argument should be a real regular expression. If it’s not the case, String::replace does exactly the same thing as String::replaceAll without the performance drawback of the regex.

Use "replace()" instead of "replaceAll()".

harshithasudhakar:
Fixes #3483 

### Motivation

When String::replaceAll is used, the first argument should be a real regular expression. If it’s not the case, String::replace does exactly the same thing as String::replaceAll without the performance drawback of the regex.

### Modifications

Replaced "replaceAll()" with "replace()"

### Documentation

 No


mxsm (approved):

Alonexc (approved):
LGTM

xwm1992 (approved):

PR #3635, opened 2023-04-06
[ISSUES #3511]Swap these 2 arguments so they are in the correct order: expected value, actual value.[ConfigurationHolderTest]

### LINKED ISSUE #3511: [Enhancement] Swap these 2 arguments so they are in the correct order: expected value, actual value.[ConfigurationHolderTest]
### Search before asking

### Enhancement Request

image
located at:
eventmesh-storage/eventmesh-storage-rabbitmq/src/test/java/org/apache/eventmesh/storage/rabbitmq/config/ConfigurationHolderTest.java line 50-60
analysis and explanation:
The standard assertions library methods such as org.junit.Assert.assertEquals, and org.junit.Assert.assertSame expect the first argument to be the expected value and the second argument to be the actual value. For AssertJ, it’s the other way around, the argument of org.assertj.core.api.Assertions.assertThat is the actual value, and the subsequent calls contain the expected values. Swap them, and your test will still have the same outcome (succeed/fail when it should) but the error messages will be confusing.

Swap actual and expected values.

harshithasudhakar:
Fixes #3511 

### Motivation

*Explain the content here.*

### Modifications

Swapped actual and expected values from lines 50-59.

### Documentation

### DIFF
--- a/eventmesh-storage-plugin/eventmesh-storage-rabbitmq/src/test/java/org/apache/eventmesh/storage/rabbitmq/config/ConfigurationHolderTest.java
+++ b/eventmesh-storage-plugin/eventmesh-storage-rabbitmq/src/test/java/org/apache/eventmesh/storage/rabbitmq/config/ConfigurationHolderTest.java
@@ -47,16 +47,16 @@ public class ConfigurationHolderTest {
     }
 
     private void assertConfig(ConfigurationHolder config) {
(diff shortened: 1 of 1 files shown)


Alonexc (approved):
LGTM

Alonexc:
image
please check checkstyle.

harshithasudhakar:
I've resolved it.

PR #3694, opened 2023-04-11
[ISSUE #3416] Used another way to initialize localClients instance [HttpClientGroupMapping]

### LINKED ISSUE #3416: [Enhancement] Use another way to initialize this instance.
### Search before asking

### Enhancement Request

image
located at:
eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/core/protocol/http/consumer/HttpClientGroupMapping.java line 391,449

analysis and explanation:
Because Double Brace Initialization (DBI) creates an anonymous class with a reference to the instance of the owning object, its use can lead to memory leaks if the anonymous inner class is returned and held by other objects. Even when there’s no leak, DBI is so obscure that it’s bound to confuse most maintainers.
For collections, use Arrays.asList instead, or explicitly add each item directly to the collection.

image

devCod3r:
Fixes #3416

### Motivation

*Because Double Brace Initialization (DBI) creates an anonymous class with a reference to the instance of the owning object, its use can lead to memory leaks if the anonymous inner class is returned and held by other objects. Even when there’s no leak, DBI is so obscure that it’s bound to confuse most maintainers.*
*Fixes the above mentioned issue.*
Issue #3416

### Modifications

*Modified the initialization of localClients instance on line 391,451.*

### Documentation

- Does this pull request introduce a new feature?
- Ans: no
- Ans: Only changed the method of initialization of an instance.

### DIFF
--- a/eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/core/protocol/http/consumer/HttpClientGroupMapping.java
+++ b/eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/core/protocol/http/consumer/HttpClientGroupMapping.java
@@ -388,11 +388,13 @@ public final class HttpClientGroupMapping {
             client.setUrl(url);
             client.setLastUpTime(new Date());
             final String groupTopicKey = client.getConsumerGroup() + "@" + client.getTopic();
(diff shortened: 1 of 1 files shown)


Alonexc (dismissed):

Alonexc:
image
Please check checkstyle.

PR #3709, opened 2023-04-12
[ISSUE #3653] refactored code and extracted method

### LINKED ISSUE #3653: [Enhancement] Code reuse is extracted as a method.[SendMessageRequestProtocolResolver]
### Search before asking

### Enhancement Request

image
located at:
eventmesh-protocol-plugin/eventmesh-protocol-meshmessage/src/main/java/org/apache/eventmesh/protocol/meshmessage/resolver/http/SendMessageRequestProtocolResolver.java line 67-98,101-132

image
image

dipankr:
…umber of line of duplicate code

Fixes #3653.

### Motivation

we have a few lines of duplicate code and the suggestion was to extract it to a method

### Modifications

refactored code to make it more readable and extracted a method

### Documentation

### DIFF
--- a/eventmesh-protocol-plugin/eventmesh-protocol-meshmessage/src/main/java/org/apache/eventmesh/protocol/meshmessage/resolver/http/SendMessageRequestProtocolResolver.java
+++ b/eventmesh-protocol-plugin/eventmesh-protocol-meshmessage/src/main/java/org/apache/eventmesh/protocol/meshmessage/resolver/http/SendMessageRequestProtocolResolver.java
@@ -59,81 +59,64 @@ public class SendMessageRequestProtocolResolver {
 
             String content = sendMessageRequestBody.getContent();
 
-            CloudEvent event = null;
             CloudEventBuilder cloudEventBuilder;
-            if (StringUtils.equals(SpecVersion.V1.toString(), protocolVersion)) {
-                cloudEventBuilder = CloudEventBuilder.v1();
 
-                cloudEventBuilder = cloudEventBuilder.withId(sendMessageRequestBody.getBizSeqNo())
-                    .withSubject(sendMessageRequestBody.getTopic())
-                    .withType("eventmeshmessage")
-                    .withSource(URI.create("/"))
-                    .withData(content.getBytes(StandardCharsets.UTF_8))
-                    .withExtension(ProtocolKey.REQUEST_CODE, code)
-                    .withExtension(ProtocolKey.ClientInstanceKey.ENV, env)
-                    .withExtension(ProtocolKey.ClientInstanceKey.IDC, idc)
-                    .withExtension(ProtocolKey.ClientInstanceKey.IP, ip)
-                    .withExtension(ProtocolKey.ClientInstanceKey.PID, pid)
-                    .withExtension(ProtocolKey.ClientInstanceKey.SYS, sys)
-                    .withExtension(ProtocolKey.ClientInstanceKey.USERNAME, username)
-                    .withExtension(ProtocolKey.ClientInstanceKey.PASSWD, passwd)
-                    .withExtension(ProtocolKey.VERSION, version.getVersion())
-                    .withExtension(ProtocolKey.LANGUAGE, language)
-                    .withExtension(ProtocolKey.PROTOCOL_TYPE, protocolType)
-                    .withExtension(ProtocolKey.PROTOCOL_DESC, protocolDesc)
-                    .withExtension(ProtocolKey.PROTOCOL_VERSION, protocolVersion)
-                    .withExtension(SendMessageRequestBody.BIZSEQNO, sendMessageRequestBody.getBizSeqNo())
-                    .withExtension(SendMessageRequestBody.UNIQUEID, sendMessageRequestBody.getUniqueId())
-                    .withExtension(SendMessageRequestBody.PRODUCERGROUP,
-                        sendMessageRequestBody.getProducerGroup())
-                    .withExtension(SendMessageRequestBody.TTL, sendMessageRequestBody.getTtl());
-                if (StringUtils.isNotEmpty(sendMessageRequestBody.getTag())) {
-                    cloudEventBuilder = cloudEventBuilder.withExtension(SendMessageRequestBody.TAG, sendMessageRequestBody.getTag());
-                }
-                if (sendMessageRequestBody.getExtFields() != null && sendMessageRequestBody.getExtFields().size() > 0) {
-                    for (Map.Entry<String, String> entry : sendMessageRequestBody.getExtFields().entrySet()) {
-                        cloudEventBuilder = cloudEventBuilder.withExtension(entry.getKey(), entry.getValue());
-                    }
-                }
-                event = cloudEventBuilder.build();
-            } else if (StringUtils.equals(SpecVersion.V03.toString(), protocolVersion)) {
-                cloudEventBuilder = CloudEventBuilder.v03();
-                cloudEventBuilder = cloudEventBuilder.withId(sendMessageRequestBody.getBizSeqNo())
-                    .withSubject(sendMessageRequestBody.getTopic())
-                    .withType("eventmeshmessage")
-                    .withSource(URI.create("/"))
-                    .withData(content.getBytes(StandardCharsets.UTF_8))
-                    .withExtension(ProtocolKey.REQUEST_CODE, code)
-                    .withExtension(ProtocolKey.ClientInstanceKey.ENV, env)
-                    .withExtension(ProtocolKey.ClientInstanceKey.IDC, idc)
-                    .withExtension(ProtocolKey.ClientInstanceKey.IP, ip)
-                    .withExtension(ProtocolKey.ClientInstanceKey.PID, pid)
-                    .withExtension(ProtocolKey.ClientInstanceKey.SYS, sys)
-                    .withExtension(ProtocolKey.ClientInstanceKey.USERNAME, username)
-                    .withExtension(ProtocolKey.ClientInstanceKey.PASSWD, passwd)
-                    .withExtension(ProtocolKey.VERSION, version.getVersion())
-                    .withExtension(ProtocolKey.LANGUAGE, language)
-                    .withExtension(ProtocolKey.PROTOCOL_TYPE, protocolType)
-                    .withExtension(ProtocolKey.PROTOCOL_DESC, protocolDesc)
-                    .withExtension(ProtocolKey.PROTOCOL_VERSION, protocolVersion)
-                    .withExtension(SendMessageRequestBody.BIZSEQNO, sendMessageRequestBody.getBizSeqNo())
-                    .withExtension(SendMessageRequestBody.UNIQUEID, sendMessageRequestBody.getUniqueId())
-                    .withExtension(SendMessageRequestBody.PRODUCERGROUP,
-                        sendMessageRequestBody.getProducerGroup())
-                    .withExtension(SendMessageRequestBody.TTL, sendMessageRequestBody.getTtl());
-                if (StringUtils.isNotEmpty(sendMessageRequestBody.getTag())) {
-                    cloudEventBuilder = cloudEventBuilder.withExtension(SendMessageRequestBody.TAG, sendMessageRequestBody.getTag());
-                }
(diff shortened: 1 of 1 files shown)


dipankr:
I was able to replace the if/else-if with a switch case and get rid of StringUtils.equals method because: 
in our case the comparisons are against SpecVersion.V1 / SpecVersion.V03,
which is not null 
so we do not require the case of StringUtils.equals(null, null) to be true as we never compare against null.

We can get away using a simple String comparison used in the switch().

for reference (source: apache commons lang doc):
Screenshot

eventmesh-protocol-plugin/eventmesh-protocol-meshmessage/src/main/java/org/apache/eventmesh/protocol/meshmessage/resolver/http/SendMessageRequestProtocolResolver.java
```diff
@@ -59,81 +59,64 @@ public static CloudEvent buildEvent(Header header, Body body) throws ProtocolHan
...
+            CloudEvent event = getBuilderCloudEvent(sendMessageRequestBody, protocolType, protocolDesc, protocolVersion,
+                code, env, idc, ip, pid, sys, username, passwd, version, language, content, cloudEventBuilder);
             return event;
+
         } catch (Exception e) {
             throw new ProtocolHandleException(e.getMessage(), e.getCause());
         }
     }
+
+    private static CloudEvent getBuilderCloudEvent(SendMessageRequestBody sendMessageRequestBody, String protocolType, String protocolDesc,
```
Alonexc, line 84: There are too many parameters in this method, maybe we can simplify it, I think it can be like this:
```
        private static CloudEvent getCloudBuilderEvent(SendMessageRequestBody sendMessageRequestBody,
                                                   SendMessageRequestHeader sendMessageRequestHeader,
                                                   CloudEventBuilder cloudEventBuilder) {
        final CloudEvent event;
        cloudEventBuilder = cloudEventBuilder.withId(sendMessageRequestBody.getBizSeqNo())
            .withSubject(sendMessageRequestBody.getTopic())
            .withType("eventmeshmessage")
            .withSource(URI.create("/"))
            .withData(sendMessageRequestBody.getContent().getBytes(StandardCharsets.UTF_8))
            .withExtension(ProtocolKey.REQUEST_CODE, sendMessageRequestHeader.getCode())
            .withExtension(ProtocolKey.ClientInstanceKey.ENV, sendMessageRequestHeader.getEnv())
            .withExtension(ProtocolKey.ClientInstanceKey.IDC, sendMessageRequestHeader.getIdc())
            .withExtension(ProtocolKey.ClientInstanceKey.IP, sendMessageRequestHeader.getIp())
            .withExtension(ProtocolKey.ClientInstanceKey.PID, sendMessageRequestHeader.getPid())
            .withExtension(ProtocolKey.ClientInstanceKey.SYS, sendMessageRequestHeader.getSys())
            .withExtension(ProtocolKey.ClientInstanceKey.USERNAME, sendMessageRequestHeader.getUsername())
            .withExtension(ProtocolKey.ClientInstanceKey.PASSWD, sendMessageRequestHeader.getPasswd())
            .withExtension(ProtocolKey.VERSION, sendMessageRequestHeader.getVersion().getVersion())
            .withExtension(ProtocolKey.LANGUAGE, sendMessageRequestHeader.getLanguage())
            .withExtension(ProtocolKey.PROTOCOL_TYPE, sendMessageRequestHeader.getProtocolType())
            .withExtension(ProtocolKey.PROTOCOL_DESC, sendMessageRequestHeader.getProtocolDesc())
            .withExtension(ProtocolKey.PROTOCOL_VERSION, sendMessageRequestHeader.getProtocolVersion())
            .withExtension(SendMessageRequestBody.BIZSEQNO, sendMessageRequestBody.getBizSeqNo())
            .withExtension(SendMessageRequestBody.UNIQUEID, sendMessageRequestBody.getUniqueId())
            .withExtension(SendMessageRequestBody.PRODUCERGROUP,
                sendMessageRequestBody.getProducerGroup())
            .withExtension(SendMessageRequestBody.TTL, sendMessageRequestBody.getTtl());
        if (StringUtils.isNotEmpty(sendMessageRequestBody.getTag())) {
            cloudEventBuilder = cloudEventBuilder.withExtension(SendMessageRequestBody.TAG, sendMessageRequestBody.getTag());
        }
        if (sendMessageRequestBody.getExtFields() != null && sendMessageRequestBody.getExtFields().size() > 0) {
            for (Map.Entry<String, String> entry : sendMessageRequestBody.getExtFields().entrySet()) {
                cloudEventBuilder = cloudEventBuilder.withExtension(entry.getKey(), entry.getValue());
            }
        }
        event = cloudEventBuilder.build();
        return event;
	}
```
What do you think?

dipankr: This makes a lot of sense, will update the PR

dipankr: updated the PR, please have a look
thanks

eventmesh-protocol-plugin/eventmesh-protocol-meshmessage/src/main/java/org/apache/eventmesh/protocol/meshmessage/resolver/http/SendMessageRequestProtocolResolver.java
```diff
@@ -42,98 +42,64 @@ public static CloudEvent buildEvent(Header header, Body body) throws ProtocolHan
...
+                    break;
+                case SpecVersion.V03.toString():
+                    cloudEventBuilder = CloudEventBuilder.v03();
+                    break;
+                default:
+                    return null; // unsupported protocol version
             }
+
+            CloudEvent event = getBuilderCloudEvent(sendMessageRequestHeader, sendMessageRequestBody, cloudEventBuilder);
             return event;
```
Alonexc, line 60: I think this could be done here:
```
return getBuilderCloudEvent(sendMessageRequestBody, sendMessageRequestHeader, cloudEventBuilder);
```
There is no need to use a redundant intermediate amount.

eventmesh-protocol-plugin/eventmesh-protocol-meshmessage/src/main/java/org/apache/eventmesh/protocol/meshmessage/resolver/http/SendMessageRequestProtocolResolver.java
```diff
@@ -42,98 +42,64 @@ public static CloudEvent buildEvent(Header header, Body body) throws ProtocolHan
...
+        if (StringUtils.isNotEmpty(sendMessageRequestBody.getTag())) {
+            cloudEventBuilder = cloudEventBuilder.withExtension(SendMessageRequestBody.TAG, sendMessageRequestBody.getTag());
+        }
+        if (sendMessageRequestBody.getExtFields() != null && sendMessageRequestBody.getExtFields().size() > 0) {
+            for (Map.Entry<String, String> entry : sendMessageRequestBody.getExtFields().entrySet()) {
+                cloudEventBuilder = cloudEventBuilder.withExtension(entry.getKey(), entry.getValue());
+            }
+        }
+        event = cloudEventBuilder.build();
+        return event;
```
Alonexc, line 102: Here also:
```
return cloudEventBuilder.build();
```

dipankr:
updated the PR

eventmesh-protocol-plugin/eventmesh-protocol-meshmessage/src/main/java/org/apache/eventmesh/protocol/meshmessage/resolver/http/SendMessageRequestProtocolResolver.java
dipankr: Agreed

dipankr:
Any suggestion on which approach I can take from the following :

1. Revert to using if/else-if
2. Declare 2 constants to hold the values of the versions

Thanks for the input.

Alonexc:
switch is doing string comparison, the condition of case must be a constant to work, and according to the underlying code, this class is of enum type.
image

If you want to use switch instead, I have a suggestion:
image

Use this method in the SpecVersion class.It is possible to not declare constants.
image

Please correct me if there are any errors.

dipankr:
> switch is doing string comparison, the condition of case must be a constant to work, and according to the underlying code, this class is of enum type.
> image
> 
> If you want to use switch instead, I have a suggestion:
> image
> 
> Use this method in the SpecVersion class.It is possible to not declare constants.
> image
> 
> Please correct me if there are any errors.
> 

This is perfect, updating the PRs

dipankr:
@Alonexc btw, i haven't been able to get the build right in my local. I tried debugging but haven't found the root cause yet.

Can we connect so that I can share more details on this? You can find me on slack as well. (dipanker_kr@outlook.com)

eventmesh-protocol-plugin/eventmesh-protocol-meshmessage/src/main/java/org/apache/eventmesh/protocol/meshmessage/resolver/http/SendMessageRequestProtocolResolver.java
```diff
@@ -42,98 +42,59 @@ public static CloudEvent buildEvent(Header header, Body body) throws ProtocolHan
...
-                    cloudEventBuilder = cloudEventBuilder.withExtension(SendMessageRequestBody.TAG, sendMessageRequestBody.getTag());
-                }
-                if (sendMessageRequestBody.getExtFields() != null && sendMessageRequestBody.getExtFields().size() > 0) {
-                    for (Map.Entry<String, String> entry : sendMessageRequestBody.getExtFields().entrySet()) {
-                        cloudEventBuilder = cloudEventBuilder.withExtension(entry.getKey(), entry.getValue());
-                    }
-                }
-                event = cloudEventBuilder.build();
+            switch (SpecVersion.parse(sendMessageRequestHeader.getProtocolVersion())) {
+                case SpecVersion.V1:
```
Alonexc, line 47: enum switch case label must be the unqualified name of an enumeration constant.
```
case V1:
         cloudEventBuilder = CloudEventBuilder.v1();
         break;
 case V03:
         cloudEventBuilder = CloudEventBuilder.v03();
         break;
```

dipankr: > enum switch case label must be the unqualified name of an enumeration constant.
> 
> ```
> case V1:
>          cloudEventBuilder = CloudEventBuilder.v1();
>          break;
>  case V03:
>          cloudEventBuilder = CloudEventBuilder.v03();
>          break;
> ```
right but that would mean we would be doing a static import of SpecVersion right?

Alonexc: This class is just an enumeration class, which is already used in the original code.

dipankr: thank you for the help, I had some confusion around the syntax. I could use some help setting up my local. Having build issues because of which I am not able to use a lot of IDE stuffs.

Alonexc: Enumeration types are defined in switch statements. For each case clause, the compiler simply checks for the presence of text in the enumeration, and if it is allowed to specify fully qualified class names in the case clause, then the compiler must perform a useless step.

xwm1992:
This issue #3653 has been fixed, I closed this pr.

PR #3744, opened 2023-04-16
[ISSUE #3002] flipped the string comparison 

### LINKED ISSUE #3002: [Enhancement] Method makes literal string comparisons passing the literal as an argument [RegistryHandler]
### Search before asking

### Enhancement Request

image
located at:
eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/admin/handler/RegistryHandler.java line 121,124
analysis and explanation:
This line is in the form of String str = str.equals("someOtherString");
A NullPointerException may occur if the String variable str is null. If instead the code was restructured to String str = "someOtherString".equals(str);that is, call equals() on the string literal, passing the variable as an argument, then this exception could never happen as both equals() check for null.

Adjust the code to String str = "someOtherString".equals(str).

dipankr:
Fixes #3002.

### Motivation
in string comparison the literal was the second arg, getting compared against an object, possibility of NPE

### Modifications
flipped the order of string, using literal to call the .equals

### Documentation

### DIFF
--- a/eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/admin/handler/RegistryHandler.java
+++ b/eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/admin/handler/RegistryHandler.java
@@ -83,7 +83,7 @@ public class RegistryHandler extends AbstractHttpHandler {
                 );
                 getRegistryResponseList.add(getRegistryResponse);
             }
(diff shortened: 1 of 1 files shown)


Alonexc (approved):
LGTM

xwm1992 (approved):

PR #3755, opened 2023-04-18
use a primitive boolean expression in GithubProtocol

### LINKED ISSUE #3720: [Enhancement] Use a primitive boolean expression here.
### Search before asking

### Enhancement Request

image
located at:
eventmesh-webhook/eventmesh-webhook-receive/src/main/java/org/apache/eventmesh/webhook/receive/protocol/GithubProtocol.java line 59
analysis and explanation:
When boxed type java.lang.Boolean is used as an expression to determine the control flow it will throw a NullPointerException if the value is null .It is safer to avoid such conversion altogether and handle the null value explicitly.Note, however, that no issues will be raised for Booleans that have already been null-checked.

image

Ceilzcx:
Fixes #3720 

### Motivation

*Explain the content here.*

### Modifications

change to use Boolean.FALSE to avoid NullPointException

### Documentation

- If yes, how is the feature documented? not applicable


Alonexc (approved):

weihubeats (approved):
LGTM

PR #3777, opened 2023-04-20
[ISSUE #3013] Updated ConfigurationHandler.java 

### LINKED ISSUE #3013: [Enhancement] Found reliance on default encoding: String.getBytes() [ConfigurationHandler]
### Search before asking

### Enhancement Request

image
located at:
eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/admin/handler/ConfigurationHandler.java line 107,108,118,119
analysis and explanation:

e.g.
image

piyush280599:
Fixes #3013


Alonexc:
Missing import:

xwm1992 (approved):

Issue #3015, opened 2023-02-02
[Enhancement] Found reliance on default encoding: String.getBytes() [GrpcClientHandler]


Alonexc:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/eventmesh/issues?q=is%3Aissue) and found no similar issues.

### Enhancement Request

![image](https://user-images.githubusercontent.com/91315508/216221020-77071659-fcdd-4786-9207-9dafec91e75a.png)
located at:
eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/admin/handler/GrpcClientHandler.java line 107,108,162,163,173,174
analysis and explanation:
Found a call to a method which will perform a byte to String (or String to byte) conversion, and will assume that the default platform encoding is suitable. This will cause the application behaviour to vary between platforms.

### Describe the solution you'd like

Use an alternative API and specify a charset name or Charset object explicitly.
e.g.
![image](https://user-images.githubusercontent.com/91315508/216492257-c2803b2d-81fa-4d46-83d0-8d43543939d9.png)
eventmesh-common/src/main/java/org/apache/eventmesh/common/Constants.java

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

PR #3779, opened 2023-04-20
[ISSUE #3015] Found reliance on default encoding: String.getBytes() [GrpcClientHandler]

piyush280599:


Alonexc:
Missing import:

xwm1992 (approved):

PR #3781, opened 2023-04-20
[ISSUE #3018] Found reliance on default encoding: String.getBytes() [RegistryHandler]

### LINKED ISSUE #3018: [Enhancement] Found reliance on default encoding: String.getBytes() [RegistryHandler]
### Search before asking

### Enhancement Request

image
located at:
eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/admin/handler/RegistryHandler.java line 90,91,95,96,106,107
analysis and explanation:

e.g.
image

piyush280599:
Fixes #3018


Alonexc:
Missing import:

eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/admin/handler/RegistryHandler.java
```diff
@@ -24,6 +24,7 @@
 import org.apache.eventmesh.runtime.admin.response.GetRegistryResponse;
 import org.apache.eventmesh.runtime.common.EventHttpHandler;
 import org.apache.eventmesh.runtime.registry.Registry;
+import org.apache.eventmesh.common.Constants;
```
Alonexc, line 27: The location of this import should be here:
image

xwm1992 (approved):

PR #3819, opened 2023-04-25
[ISSUE #3815]DelayRetryable and RetryContext can be refactored.

### LINKED ISSUE #3815: [Enhancement] DelayRetryable and RetryContext can be refactored.
### Search before asking

### Enhancement Request

There are 3 sub packages (.grpc, .http, .tcp) in package runtime.core.protocol. 

1. 3 `RetryContext` in 3 packages have almost the same structure. `DelayRetryable`‘s situation is similar.

2. https://github.com/apache/eventmesh/blob/master/eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/core/protocol/grpc/retry/RetryContext.java#L30 

This logic in  `RetryContext` constructor does not calculate the num of delays. The same problem as http/`RetryContext`.

3. A field in class has its own default value.

1. 3 `RetryContext` in 3 packages can be extracted and abstracted to 1  `RetryContext`.
Interface `DelayRetryable` in .grpc and .http can be extracted to one. For reusability, the interface `DelayRetryable` in .tcp can be divided into `DelayRetryable` + `RetryContext`.

2. https://github.com/apache/eventmesh/blob/master/eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/core/protocol/grpc/retry/RetryContext.java#L30 
The num of delays should be considered.

3. Remove unnecessary default value of a member field.

pandaapo:
Fixes #3815.

### Motivation

*See "Enhancement Request" in the [issue](https://github.com/apache/eventmesh/issues/3815#issue-1682617238).*

### Modifications

*See "Describe the solution you'd like" in the [issue](https://github.com/apache/eventmesh/issues/3815#issue-1682617238)*

### Documentation

- Does this pull request introduce a new feature?  no


eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/core/protocol/http/producer/SendMessageContext.java
```diff
@@ -123,13 +123,13 @@ public String toString() {
     }
 
     @Override
-    public boolean retry() throws Exception {
+    public void retry() throws Exception {
         if (eventMeshProducer == null) {
-            return false;
+            log.error("Exception happends during retry. EventMeshProduceer is null.");
```
xwm1992, line 128: here need add return in if block

eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/core/protocol/http/producer/SendMessageContext.java
```diff
@@ -123,13 +123,13 @@ public String toString() {
...
         }
 
         if (retryTimes > 0) { //retry once
-            return false;
+            log.error("Exception happends during retry. The retryTimes > 0.");
```
xwm1992, line 132: same as above

xwm1992 (changes requested):

eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/core/protocol/http/producer/SendMessageContext.java
pandaapo: I was merging 3 `RetryContext`s in module http, grpc and tcp to one.

3 `RetryContext`s have almost same structure except a little difference. This `retry()` is that difference. It returns boolean in module http, grpc, and void in module tcp. Then I checked all calls of `retry()` that returns boolean. I found the return value of boolean has no practical use in all calls. So I chose to modify it to `void` to keep with `retry()` in module tcp.

If the return value of boolean type is indeed useful, may I modify the void to boolean in module tcp?

xwm1992: What I mean is that in your modified logic, the return is missing in the if code block, and return boolean is not required.
you need add `return` after the `log.error()`.

pandaapo: Oh! Yes. :joy:

xwm1992 (approved):

PR #3845, opened 2023-04-28
[ISSUE #3823]Math operands should be cast before assignment.

### LINKED ISSUE #3823: [Enhancement] Math operands should be cast before assignment.
### Search before asking

### Enhancement Request

image
located at:
eventmesh-storage-plugin/eventmesh-storage-standalone/src/main/java/org/apache/eventmesh/storage/standalone/admin/StandaloneAdmin.java line 80
analysis and explanation:
When arithmetic is performed on integers, the result will always be an integer. You can assign that result to a long, double, or float with automatic type conversion, but having started as an int or long, the result will likely not be what you expect.
In either case, the result will not be what was expected. Instead, at least one operand should be cast or promoted to the final type before the operation takes place.

image

nikam14:
Fixes #3823 .

### Motivation

code optimisation.

### Modifications

stored the calculated value in variable then passed in to the object .

### Documentation


Alonexc (approved):
LGTM

xwm1992 (approved):

PR #3860, opened 2023-05-02
[ISSUE #3829]Do some code optimization.[CloudEventUtils]

### LINKED ISSUE #3829: [Enhancement] Do some code optimization.[CloudEventUtils]
### Search before asking

### Enhancement Request

image
image
located at:
eventmesh-storage-plugin/eventmesh-storage-rocketmq/src/main/java/org/apache/eventmesh/storage/rocketmq/utils/CloudEventUtils.java line  88,128
analysis and explanation:
a. The underlying implementation of String::replaceAll calls the java.util.regex.Pattern.compile() method each time it is called even if the first argument is not a regular expression. This has a significant performance cost and therefore should be used with care.
When String::replaceAll is used, the first argument should be a real regular expression. If it’s not the case, String::replace does exactly the same thing as String::replaceAll without the performance drawback of the regex.
b. This method prints a stack trace to the console. This is non configurable, and causes an application to look unprofessional.

a.Replace with replace().
image

b. Remove e.printStackTrace().

harshithasudhakar:
Fixes #3829 

### Modifications

Replaced `replaceAll()` with `replace()` at line 88
Removed `e.printStackTrace()` at line 128

### Documentation


mxsm (approved):
LGTM~

Alonexc (approved):

mroccyen (approved):
LGTM

PR #3867, opened 2023-05-04
[ISSUE #3861]Not yet make full use of Polymorphism in Producer and AbstractProducer.

### LINKED ISSUE #3861: [Enhancement] Not yet make full use of Polymorphism in Producer and AbstractProducer.
### Search before asking

### Enhancement Request

* In storage modules of knative, rocketmq, pulsar, kafka, the model of `Producer`'s hierarchy is not unified and confused: 3 `AbstractProducer`s in module knative, rocketmq, pulsar have some similar contents, but some implement `Producer` and some not. Even `AbstractProducer` in kafka is not been designed into `Producer`'s hierarchy.
* `InterruptedException` should never be ignored. Or the import interrupted state of Thread may be ignored unexpected.
* Some logic has been provided by existing util, so no need to code again.
* Smaller `synchronized` scope, lower the thread competitiveness
* In AtomicXxx, `set()` is thread safe and better than CAS when no need to get a return value.

* Unify the model of `Producer`'s hierarchy: Rename 3 `AbstractProducer` in module knative, rocketmq to `KnativeAbstractProducer, RocketmqAbstractProducer`. Add a `AbstractProducer` in api module implements `Producer`. Make `AbstractProducer`s in module knative, kafka, rocketmq, pulsar all extends `AbstractProducer` in api module or its sub abstract class. Remove `AbstractProducer` in module pulsar.
* Catch `InterruptedException` and call `interrupt()` of current Thread.
* Use existing util.
* Minimize `synchronized` scope.
* Use `set` replace CAS where no need get a return value.

pandaapo:
Fixes #3861.

### Motivation

See "Enhancement Request" in [issue](https://github.com/apache/eventmesh/issues/3861#issue-1694155912)

### Modifications

See "Describe the solution you'd like" in [issue](https://github.com/apache/eventmesh/issues/3861#issue-1694155912)

### Documentation


mroccyen:
@pandaapo Please fixes conflict.

eventmesh-storage-plugin/eventmesh-storage-kafka/src/main/java/org/apache/eventmesh/storage/kafka/producer/ProducerImpl.java
```diff
@@ -42,36 +41,20 @@
...
-public class ProducerImpl {
+public class ProducerImpl extends AbstractProducer {
 
     private final KafkaProducer<String, CloudEvent> producer;
-    private final Properties properties = new Properties();
-    private final AtomicBoolean isStarted = new AtomicBoolean(false);
 
     public ProducerImpl(Properties props) {
+        super(props);
+        properties.clear();
```
xwm1992, line 50: why use `clear()` method at here

xwm1992 (changes requested):

eventmesh-storage-plugin/eventmesh-storage-kafka/src/main/java/org/apache/eventmesh/storage/kafka/producer/ProducerImpl.java
pandaapo: Because `super(props)` assigns param `props` to field `properties`. but `ProducerImpl` does not assign value like that. What it do to `properties` is just the next 3 line.
https://github.com/apache/eventmesh/blob/470329d7e3c69fa241762b364f1cd1799a3b5766/eventmesh-storage-plugin/eventmesh-storage-kafka/src/main/java/org/apache/eventmesh/storage/kafka/producer/ProducerImpl.java#L51-L54

Pil0tXia: Excuse me but I still don't understand that why Kafka needs a `clear()` statement here but RocketMQ and Pulsar don't.

pandaapo: Then I explain it in detail again.

The following is the original logic of this method under Kafka module. You can see that only three items need to be set in `properties` object. Now after the unified abstraction, `properties` is first assigned a value, which is equivalent to setting more items. If I don't `clear()` first, it firstly disrupts the original logic, secondly causes data contamination, and thirdly, the additional unknown items set might lead to unnecessary and unknown problems.

https://github.com/apache/eventmesh/blob/9ce5dd6fd94d60d47358721a18eff0f50cd932ad/eventmesh-storage-plugin/eventmesh-storage-kafka/src/main/java/org/apache/eventmesh/storage/kafka/producer/ProducerImpl.java#L51-L56

As for why RocketMQ and Pulsar modules don't need `clear()`, because it is consistent with their original logic.

Pil0tXia: Then how about moving the `properties.put` statements of `org.apache.eventmesh.storage.kafka.producer.ProducerImpl` into the `init()` method of `org.apache.eventmesh.storage.kafka.producer.KafkaProducerImpl`?

pandaapo: Actually, when we read the code of `ProducerImpl` and `KafkaProducerImpl` , it is quite evident that the latter is merely a delegate class, delegating all the specific details to former for actual processing. Of course, you can use the code from the corresponding classes in RocketMQ module to refute me. However, I personally favor the positioning of these two classes in Kafka module, as their responsibilities are more clearly defined.

Finally Community Over Code. I'm willing to wait for more members in community to express their views on this point. If the approach used in RocketMQ module is more accepted, I will revise.

Pil0tXia: >it is quite evident that the latter is merely a delegate class, delegating all the specific details to former for actual processing

I agree with your view. Since the abstraction has already been made, how about standardizing the usage of the abstract implementation?

pandaapo: My view on what you call 'standardizing' is similar to https://github.com/apache/eventmesh/pull/3867#discussion_r1576133454 above.

PR #3969, opened 2023-05-18
[ISSUE #3964] Used Exception instead of Throwable

### LINKED ISSUE #3964: [Enhancement] Catch Exception instead of Throwable.
### Search before asking

### Enhancement Request

image
located at:
eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/trace/TraceUtils.java line 45,58,77
analysis and explanation:
Throwable is the superclass of all errors and exceptions in Java. Error is the superclass of all errors, which are not meant to be caught by applications.
Catching either Throwable or Error will also catch OutOfMemoryError and InternalError, from which an application should not attempt to recover.

Replace with `Exception`.
image

ZahaanMahajan:
Fixes #3964


Alonexc:
LGTM.
image
Need to link the issue in the description, example: Fixes #XXXX

PR #3988, opened 2023-05-21
[ISSUE #481]Use YAML file to store configuration

### LINKED ISSUE #481: Use YAML file to store configuration
## Enhancement Request
Currently, there are a lot of configuration fields in EventMesh, and it will increase in the future, you can look up at `eventmesh.properties`.
Such as:
```properties
eventMesh.server.idc=DEFAULT
eventMesh.server.env=PRD
eventMesh.server.cluster=COMMON
eventMesh.server.name=EVENTMESH-runtime
eventMesh.sysid=0000
eventMesh.server.http.port=10105
########################## eventMesh tcp configuration ############################
eventMesh.server.tcp.enabled=true
eventMesh.server.tcp.port=10000
eventMesh.server.tcp.readerIdleSeconds=120
eventMesh.server.tcp.writerIdleSeconds=120
eventMesh.server.tcp.allIdleSeconds=120
eventMesh.server.tcp.clientMaxNum=10000
# client isolation time if the message send failure
eventMesh.server.tcp.pushFailIsolateTimeInMills=30000
```
It is difficult to find out which is your want since most of them has the same prefix.
It is better to store them at yaml file, the advantage is that you can merge the common prefixes like below:
```yml
eventMesh:
  server:
    idc: DEFAULT
    env: PRD
    cluster: COMMON
    name: EVENTMESH-runtime
    http:
      port: 10105
    tcp:
      enabled: true
      port: 10000
      readerIdleSeconds: 120
      writerIdleSeconds: 120
      allIdleSeconds: 120
      clientMaxNum: 10000
      pushFailIsolateTimeInMills: 30000 # client isolation time if the message send failure
  sysid: 0000

```

pandaapo:
Fixes #481.

### Motivation

See "Enhancement Request" in [issue](https://github.com/apache/eventmesh/issues/481#issue-959081252).

### Modifications

The default config file format keeps `eventmesh.properties`. When user set only config file in yaml format, EventMesh will find it and load it.

### Documentation


mxsm:
Hi @pandaapo  This feature has been implemented. You can refer to the specific class org.apache.eventmesh.common.config.FileLoad.

pandaapo:
> Hi @pandaapo This feature has been implemented. You can refer to the specific class org.apache.eventmesh.common.config.FileLoad.

Thanks for your review.
* You can try its method `FileLoad.YamlFileLoad#getConfig(ConfigInfo)`. Then will can get properties like `aaa ->, bbb -> , ccc -> 3` when the yml content is
```
aaa
  bbb
    ccc: 3
```

* Current code will never load eventmesh.yaml. Because the call stack logic in starting up with standalone mode always use  `PropertiesFileLoad`.

eventmesh-runtime/conf/eventmesh.yml
```diff
@@ -0,0 +1,141 @@
...
+    fileMode:
+      # The file storage path of the file storage mode. If ${eventmeshhome} is written, it is in the eventmesh root directory
+      filePath: ${eventMeshHome}/webhook
+    # Nacos storage mode, and the configuration naming rule is eventmesh webHook. nacosMode. {nacos native configuration key} please see the specific configuration [nacos github api](https://github.com/alibaba/nacos/blob/develop/api/src/main/java/com/alibaba/nacos/api/SystemPropertyKeyConst.java)
+    ## Address of Nacos
+    nacosMode:
+      serverAddr: 127.0.0.1:8848
+    # Webhook eventcloud sending mode. And eventmesh connector. plugin. The type configuration is the same
+    producer:
+      storage: standalone
```
Pil0tXia, line 141: image

As is seen in https://github.com/apache/eventmesh/issues/4185, According to the original Chinese comments in the git commit history, the existing English comments are misleading. The English translation does not match the meaning of the original Chinese text, which affects understanding.

```suggestion
    # Webhook CloudEvent sending mode. This property is the same as the eventMesh.storage.plugin.type configuration.
    producer:
      storage: standalone
```

pandaapo: Thanks

Pil0tXia:
I think we should only use one configuration file, either `eventmesh.yml` or `eventmesh.properties`. Having both concurrently would pose difficulties in maintenance.

Considering that connectors all use YAML configuration files, the presence of `eventmesh.properties` feels inconsistent, since it is one of the few properties files in EventMesh. Therefore, my preference would be to standardize the use of YAML files.

In our project, we currently have both `.yml` and `.yaml` files, and both file extensions are valid. However, it would be necessary to choose one for consistency. Although `.yaml` is a more standard and compliant file extension compared to `.yml`, the majority of our project uses `.yml`. Hence, I recommend standardizing on the `.yml` extension (except for `.asf.yaml`).

pandaapo:
> I think we should only use one configuration file, either `eventmesh.yml` or `eventmesh.properties`. Having both concurrently would pose difficulties in maintenance.
> 
> Considering that connectors all use YAML configuration files, the presence of `eventmesh.properties` feels inconsistent, since it is one of the few properties files in EventMesh. Therefore, my preference would be to standardize the use of YAML files.
> 
> In our project, we currently have both `.yml` and `.yaml` files, and both file extensions are valid. However, it would be necessary to choose one for consistency. Although `.yaml` is a more standard and compliant file extension compared to `.yml`, the majority of our project uses `.yml`. Hence, I recommend standardizing on the `.yml` extension (except for `.asf.yaml`).

Thanks for your review. But sorry, many of your points I couldn't agree with, and they were even a bit strange to me. No offense, I'm willing to wait for more members of the community to express their views on these points of yours. If there are no objections, I will revise.

Pil0tXia:
Any progress we can make here?

PR #4145, opened 2023-06-20
[ISSUE #4144] Subscription is almost impossible to be cancelled when the TCP sub client is closed 

### LINKED ISSUE #4144: [Bug] Subscription is almost impossible to be cancelled when the TCP sub client is closed
### Search before asking

### Environment

Windows

### EventMesh version

master

### What happened

When the TCP subscriber is closed, eventmesh-runtime is still subscribing to related topics.

### How to reproduce

Use `org.apache.eventmesh.tcp.demo.pub.cloudevents.AsyncPublish` in `eventmesh-examples` to push messages.

Use `org.apache.eventmesh.tcp.demo.sub.cloudevents.AsyncSubscribe` in `eventmesh-examples` to subscribe the topic. After message are consumed successfully, close `AsyncSubscribe`.

Use `org.apache.eventmesh.tcp.demo.pub.cloudevents.AsyncPublish` in `eventmesh-examples` to push messages again. Then some messages will still be consumed. This should not happen.

### Debug logs

_No response_

pandaapo:
Fixes #4144.

### Motivation

Subscription is almost impossible to be cancelled when the TCP sub client is closed.

### Modifications

https://github.com/apache/eventmesh/blob/8843df2f78fffaf49815403a579b810c64c1f2d6/eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/core/protocol/tcp/client/group/ClientGroupWrapper.java#L246
The subscription session can't be removed from `ClientGroupWrapper.topic2sessionInGroupMapping` correctly. 

When `Session` is added to the `topic2sessionInGroupMapping`, a `java.util.Set`, `Session.sessionState` is `CREATED`. But session is almost impossible to keep this state. It will be `RUNNING` or `CLOSED`. So the hash of `Session` is changed. `Map`(the base of `Set`) will be unable to find the bucket using the new hash, which is different from origin hash used when initially putting into `Map`.

### Documentation


lrhkobe:
Thank you for your problem. I have checked this issue, because the reference is stored in the set, so there is no problem. You can run and verify it.

pandaapo:
@lrhkobe Thank you for your review. 

`Set` indeed stores `Session`'s reference. But it does not mean `Set<Session>#remove(Session)` can remove a stored `Session` successfully. Because `Session` overrides the `hashCode()` method, it results in the hash value of `Session` will be changed after it is stored in `Set`. Just like what I explain in https://github.com/apache/eventmesh/pull/4145#issue-1765742977
>When `Session` is added to the `topic2sessionInGroupMapping`, a `java.util.Set`, `Session.sessionState` is `CREATED`. But session is almost impossible to keep this state. It will be `RUNNING` or `CLOSED`. So the hash of `Session` is changed. 

`Set#remove` uses `Map#remove` to implement. `Map` use the hash value to find an element to remove. Because the hash value is changed after `Session` is put into `Map`'s bucket, `Map#remove` will failed.

xwm1992 (approved):

lrhkobe (approved):
LGTM

PR #4244, opened 2023-07-18
[ISSUE #4200]Do some code optimization.[ServiceUtils]

### LINKED ISSUE #4200: [Enhancement] Do some code optimization.[ServiceUtils]
### Search before asking

### Enhancement Request

image
image
located at:
eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/core/protocol/grpc/service/ServiceUtils.java line 89,92,95,147
analysis and explanation:
a. This code seems to be using non-short-circuit logic (e.g., & or |) rather than short-circuit logic (&& or ||). In addition, it seem possible that, depending on the value of the left hand side, you might not want to evaluate the right hand side (because it would have side effects, could cause an exception or could be expensive.
Non-short-circuit logic causes both sides of the expression to be evaluated even when the result can be inferred from knowing the left-hand side. This can be less efficient and can result in errors if the left-hand side guards cases when evaluating the right-hand side can generate an error.
b. Dereference of 'heartbeatItems' may produce 'NullPointerException'

a. Change to use `&&`.
reference.
image
b. use Objects.requireNonNull() to avoid NPE.

VL-037:
Fixes #4200 .

### Motivation

his changes resolves #4200  by using short-circuit-logic & add `Objects.requireNonNull()`

### Modifications

`eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/core/protocol/grpc/service/ServiceUtils.java` line 89, 92, 95, 147

### Documentation

- If yes, how is the feature documented? -
- If a feature is not applicable for documentation, explain why? -

### DIFF
--- a/eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/core/protocol/grpc/service/ServiceUtils.java
+++ b/eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/core/protocol/grpc/service/ServiceUtils.java
@@ -42,6 +42,7 @@ import java.time.OffsetDateTime;
 import java.time.ZoneId;
 import java.time.ZoneOffset;
 import java.util.List;
+import java.util.Objects;
 
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.google.protobuf.Any;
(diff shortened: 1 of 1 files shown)


mxsm (approved):
LGTM

Pil0tXia (approved):
LGTM

Alonexc (approved):

xwm1992 (approved):

lrhkobe (approved):
LGTM

Issue #4252, opened 2023-07-19
[Enhancement] Typo correction in [AssertUtils] class


HattoriHenzo:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/eventmesh/issues?q=is%3Aissue) and found no similar issues.

### Enhancement Request

Correction of minor typo in the `AssertUtils `class

### Describe the solution you'd like

Rename some methods and change some comments

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #4253, opened 2023-07-19
[ISSUE #4252] Minor refactoring for AssertUtils class

HattoriHenzo:
Fixes https://github.com/apache/eventmesh/issues/4252

### Motivation
Minor refactoring on `AssertUtils ` class

### DIFF
--- a/eventmesh-common/src/main/java/org/apache/eventmesh/common/utils/AssertUtils.java
+++ b/eventmesh-common/src/main/java/org/apache/eventmesh/common/utils/AssertUtils.java
@@ -27,35 +27,35 @@ import java.util.Objects;
 public final class AssertUtils {
 
     /**
-     * Assert actual is not null
+     * Assert obj is not null
      *
-     * @param actual  Object to test
+     * @param obj  Object to test
      * @param message error message
      */
-    public static void notNull(final Object actual, final String message) {
-        isTrue(Objects.nonNull(actual), message);
+    public static void notNull(final Object obj, final String message) {
+        isTrue(Objects.nonNull(obj), message);
     }
 
     /**
-     * Assert actual is true
+     * Assert condition is true
      *
-     * @param actual  boolean to test
+     * @param condition  boolean to test
      * @param message error message
      */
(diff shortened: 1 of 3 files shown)


eventmesh-common/src/main/java/org/apache/eventmesh/common/utils/AssertUtils.java
```diff
@@ -27,35 +27,35 @@
...
      *
-     * @param test    test
+     * @param actual  boolean to test
      * @param message error message
      */
-    public static void isTrue(final Boolean test, final String message) {
-        if (!Boolean.TRUE.equals(test)) {
+    public static void isTrue(final Boolean actual, final String message) {
+        if (!Boolean.TRUE.equals(actual)) {
             throw new IllegalArgumentException(message);
```
Pil0tXia, line 47: If there is no logic in the code that needs to compare the expected value to the actual value, and there is no expected value that needs to be compared, then there is no need to name a single parameter `actual`

Besides, "actual" is an adjective.

The same applies to the other two methods.

Pil0tXia (changes requested):

eventmesh-common/src/main/java/org/apache/eventmesh/common/utils/AssertUtils.java
HattoriHenzo: @Pil0tXia Thanks for your feedback. As the idea of the test is an assertion, naming the variable as `actual `is like naming it `current`. 
Anyway after doing some [research ](https://docs.oracle.com/middleware/11119/jdev/api-reference-esdk/oracle/ide/util/Assert.html#check_boolean_) I saw that for booleans, the variable is usually named `condition`. If this LGTY, I will rename the boolean variable as `condition `and rollback the remaining variables.

Pil0tXia: Yes, `condition` is better. Good idea.

Alonexc (approved):

Issue #4288, opened 2023-07-26
[Enhancement] Enable the nacos plugin to support namespaces


1404369980:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/eventmesh/issues?q=is%3Aissue) and found no similar issues.

### Enhancement Request

Currently, the registry plugin Nacos only support to default namespace.

### Describe the solution you'd like

Make registry plugin Nacos namespace configurable .

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #4289, opened 2023-07-26
[ISSUE #4288] Enable the nacos plugin to support namespace

1404369980:
[ISSUE #4288] Enable the nacos plugin to support namespace


mxsm (commented):
@1404369980  Please link this pr to issue use fix or close.

PR #4297, opened 2023-07-27
[ISSUE #4263] InterruptedExceptions should never be ignored[ThreadUtils]

### LINKED ISSUE #4263: [Enhancement] InterruptedExceptions should never be ignored[ThreadUtils].
### Search before asking

### Enhancement Request

image
located at:
eventmesh-common/src/main/java/org/apache/eventmesh/common/utils/ThreadUtils.java line 36,53
analysis and explanation:

Restore interrupted state
image

harshithasudhakar:
Fixes #4263 

### Modifications

Restored interrupted states in lines 37 & 54

### Documentation

### DIFF
--- a/eventmesh-common/src/main/java/org/apache/eventmesh/common/utils/ThreadUtils.java
+++ b/eventmesh-common/src/main/java/org/apache/eventmesh/common/utils/ThreadUtils.java
@@ -34,7 +34,7 @@ public class ThreadUtils {
             long timeout = ThreadLocalRandom.current().nextLong(min, max + 1);
             timeUnit.sleep(timeout);
         } catch (InterruptedException ignore) {
-            Thread.currentThread().interrupt;
+            Thread.currentThread().interrupt();
         }
     }
 
@@ -51,7 +51,7 @@ public class ThreadUtils {
         try {
             sleepWithThrowException(timeout, timeUnit);
         } catch (InterruptedException ignore) {
-            Thread.currentThread().interrupt;
+            Thread.currentThread().interrupt();
         }
     }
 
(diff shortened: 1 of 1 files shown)


eventmesh-common/src/main/java/org/apache/eventmesh/common/utils/ThreadUtils.java
```diff
@@ -34,7 +34,7 @@ public static void randomPause(long min, long max, TimeUnit timeUnit) {
             long timeout = ThreadLocalRandom.current().nextLong(min, max + 1);
             timeUnit.sleep(timeout);
         } catch (InterruptedException ignore) {
-            //ignore
+            Thread.currentThread().interrupt;
```
Alonexc, line 37:
```
Thread.currentThread().interrupt();
```

eventmesh-common/src/main/java/org/apache/eventmesh/common/utils/ThreadUtils.java
```diff
@@ -51,7 +51,7 @@ public static void sleep(long timeout, TimeUnit timeUnit) {
         try {
             sleepWithThrowException(timeout, timeUnit);
         } catch (InterruptedException ignore) {
-            //ignore
+            Thread.currentThread().interrupt;
```
Alonexc, line 54: same as above.

Pil0tXia (approved):
LGTM

mxsm (approved):
LGTM

Alonexc (approved):

xwm1992 (approved):

PR #4314, opened 2023-08-01
[ISSUE #4266]InterruptedExceptions should never be ignored[connectors]

### LINKED ISSUE #4266: [Enhancement] InterruptedExceptions should never be ignored[connectors]
### Search before asking

### Enhancement Request

located at:
eventmesh-connectors/eventmesh-connector-openfunction/src/main/java/org/apache/eventmesh/connector/openfunction/sink/connector/OpenFunctionSinkConnector.java line 77
eventmesh-connectors/eventmesh-connector-openfunction/src/main/java/org/apache/eventmesh/connector/openfunction/source/connector/OpenFunctionSourceConnector.java line 90
eventmesh-connectors/eventmesh-connector-rocketmq/src/main/java/org/apache/eventmesh/connector/rocketmq/sink/connector/RocketMQSinkConnector.java 80
analysis and explanation:

Restore interrupted state
image

Ruhshan:
Fixes #4266 .

### Motivation

Interrupted exception should never be ignored in the code

### Modifications

Restored interrupted state for interrupted exception in the mentioned files in the issue.

### Documentation

- Does this pull request introduce a new feature? (yes / no) no


Alonexc:
I think log.error() should be used instead of log.warn().

xwm1992 (approved):

mxsm (approved):

PR #4334, opened 2023-08-05
[ISSUE #4264][Enhancement] Some optimizations for ProducerService

### LINKED ISSUE #4264: [Enhancement] Some optimizations for ProducerService.
### Search before asking

### Enhancement Request

image
image
located at:
eventmesh-connectors/eventmesh-connector-openfunction/src/main/java/org/apache/eventmesh/connector/openfunction/service/ProducerService.java
analysis and explanation:
a. and has three main disadvantages:
Additional behavior such as validation cannot be added.
The internal representation is exposed, and cannot be changed afterwards.
Member values are subject to change from anywhere in the code and may not meet the programmer’s assumptions.
By using private attributes and accessor methods (set and get), unauthorized modifications are prevented.
b. InterruptedExceptions should never be ignored in the code, and simply logging the exception counts in this case as "ignoring". The throwing of the InterruptedException clears the interrupted state of the Thread, so if the exception is not handled properly the information that the thread was interrupted will be lost. Instead, InterruptedExceptions should either be rethrown - immediately or after cleaning up the method’s state - or the thread should be re-interrupted by calling Thread.interrupt() even if this is supposed to be a single-threaded application. Any other course of action risks delaying thread shutdown and loses the information that the thread was interrupted - probably without finishing its task.

a. Make field non-public and provide accessors if needed.
image
b. Restore interrupted state
image

847850277:
Fixes #4264.

1、Make field non-public and provide accessors if needed.
2、Restore interrupted state

### DIFF
--- a/eventmesh-connectors/eventmesh-connector-openfunction/src/main/java/org/apache/eventmesh/connector/openfunction/service/ConsumerService.java
+++ b/eventmesh-connectors/eventmesh-connector-openfunction/src/main/java/org/apache/eventmesh/connector/openfunction/service/ConsumerService.java
@@ -42,11 +42,11 @@ import lombok.extern.slf4j.Slf4j;
 @Slf4j
 public class ConsumerService extends ConsumerServiceGrpc.ConsumerServiceImplBase {
 
-    public OpenFunctionSinkConnector openFunctionSinkConnector;
+    private final OpenFunctionSinkConnector openFunctionSinkConnector;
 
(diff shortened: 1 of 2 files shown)


Pil0tXia (approved):
LGTM

mxsm:
@847850277 thanks for your contribution

mxsm (approved):
LGTM

PR #4364, opened 2023-08-15
[ISSUE#4095] Code Optimization and Interrupeted Exception handling.[EtcdCustomService]

### LINKED ISSUE #4095: [Enhancement] Code optimization. [EtcdCustomService]
### Search before asking

### Enhancement Request

image
image
image
analysis and explanation:
a. InterruptedExceptions should never be ignored in the code, and simply logging the exception counts in this case as "ignoring". The throwing of the InterruptedException clears the interrupted state of the Thread, so if the exception is not handled properly the information that the thread was interrupted will be lost. Instead, InterruptedExceptions should either be rethrown - immediately or after cleaning up the method’s state - or the thread should be re-interrupted by calling Thread.interrupt() even if this is supposed to be a single-threaded application. Any other course of action risks delaying thread shutdown and loses the information that the thread was interrupted - probably without finishing its task.
b. Declaring a variable only to immediately return or throw it is a bad practice.
Some developers argue that the practice improves code readability, because it enables them to explicitly name what is being returned. However, this variable is an internal implementation detail that is not exposed to the callers of the method. The method name should be sufficient for callers to know exactly what will be returned.

a. restore interrupted state
image

b. inline variable
image

himansh295:
Fixes #4095 .

### Motivation

*Explain the content here.*

### Modifications

### Documentation

### DIFF
--- a/eventmesh-registry-plugin/eventmesh-registry-etcd/src/main/java/org/apache/eventmesh/registry/etcd/service/EtcdCustomService.java
+++ b/eventmesh-registry-plugin/eventmesh-registry-etcd/src/main/java/org/apache/eventmesh/registry/etcd/service/EtcdCustomService.java
@@ -94,10 +94,10 @@ public class EtcdCustomService extends EtcdRegistryService {
                 );
             }
         } catch (Exception e) {
+            log.error("[EtcdRegistryService][findEventMeshAppSubTopicInfoByGroup] error, group: {}", group, e);
             if(e instanceof InterruptedException) {
                 Thread.currentThread().interrupt();
             }
-            log.error("[EtcdRegistryService][findEventMeshAppSubTopicInfoByGroup] error, group: {}", group, e);
             throw new RegistryException(e.getMessage());
         }
         return null;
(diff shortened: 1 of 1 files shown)


eventmesh-registry-plugin/eventmesh-registry-etcd/src/main/java/org/apache/eventmesh/registry/etcd/service/EtcdCustomService.java
```diff
@@ -68,6 +68,9 @@ public List<EventMeshServicePubTopicInfo> findEventMeshServicePubTopicInfos() th
                 return eventMeshServicePubTopicInfoList;
             }
         } catch (Exception e) {
+            if(e instanceof InterruptedException) {
```
Alonexc, line 71: I think it would be better to put it after the log message.

himansh295: ohk sure

mxsm (commented):
@himansh295  Please fix code style
image

Pil0tXia:
Please resolve conflicts.

Pil0tXia:
Hi @himansh295 , the issue has been accomplished in https://github.com/apache/eventmesh/pull/4752. Do you have any additional information to share? Feel free to share your latest insights with us. If you feel that your issue has been resolved, you are welcome to close it. We sincerely appreciate your contributions to Apache EventMesh.

Issue #4043, opened 2023-05-29
[GLCC] GitLink Code Camp (CCF GitLink 开源编程夏令营)


xwm1992:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/eventmesh/issues?q=is%3Aissue) and found no similar issues.

## 题目名称
filter与transformer设计与插件化实现

### 题目简介
filter与transformer是eventmesh的事件过滤与事件转换组件，支持插件化设计，在eventmesh收发事件过程中可配置事件过滤与转换的规则，实现过滤与转换的能力。

### 编码任务
包括但不限于：
- 熟悉现有eventmesh功能
- 设计编码eventmesh-filter模块
- 设计编码eventmesh- transformer模块
- 输出对应使用demo与文档介绍

### 技能要求和编程语言
- Java
- 对eventmesh有一定的了解
- 对k8s有一定的了解

### 预期成果
- 支持eventmesh-filter配置不同过滤规则进行事件收发
- 支持eventmesh-transformer配置不同转换规则进行事件收发
- 输出相关操作指引与使用文档

### 可参与人数
1人（项目由个人独立完成）

### 导师及联系方式
GitHub ID：qqeasonchen
Email：chenguangsheng@apache.org

### 官方微信群
![wechat-assistant](https://github.com/apache/eventmesh/assets/13237619/d4582ab1-948c-4271-806f-1438bd2eb6c4)

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

KeeProMise:
Hi~ I'm interested in this task, hope to get in touch with you!

993526915:
I`m very interested in this project,hope you can give me this opportunity to complete this project！

Alonexc:
Welcome！

ShockleysxX:
Hi, @qqeasonchen, I am very interested in this project, can I still apply?

Issue #3263, opened 2023-02-24
[Feature] The filtering feature of events 


epiao55:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/eventmesh/issues?q=is%3Aissue) and found no similar issues.

### Feature Request

I would like to contribute a filtering feature for events. Here is my rough design idea.
First, when creating a subscription, add a new json String field 'filter' to the SubscriptionItem to describe the specific filtering rules.for example
```json
{
	"and": [{
		"field": "specversion",
		"operator": "EQ",
		"value": "1.0"
	}, {
		"or": [{
			"field": "id",
			"operator": "EQ",
			"value": "9aeb0fdf-c01e-0131-0922-9eb54906e209"
		}, {
			"field": "email",
			"operator": "CN",
			"value": "example"
		}]
	}, {
		"field": "data.contentType",
		"operator": "SW",
		"value": "image/png"
	}]
}
```
When a consumer consumes, check the 'filter' field and convert it into a 'FilterGroup'. Then, generate a combined predicate based on the rules and provide the filtering results.
The logical relationship between the rules is described by 'and' and 'or'. The 'field' field describes the field to be checked, the 'operator' describes the specific rule, and the 'value' describes the comparison data for the rule.
The value of the field field can be a nested field in the event, expressed as a hierarchical relationship using  '.'.
and This is the filtering process.
```mermaid
sequenceDiagram
    participant Consumer
    participant FilterGroup
    participant SubscriptionItem
    participant SubscriptionItem
    Consumer->>+FilterGroup: Transform event to JSON
    FilterGroup->>SubscriptionItem: Get filter rules JSON
    FilterGroup->>FilterGroup: Parse filter rules JSON
    FilterGroup->>FilterGroup: Generate combined predicate
    FilterGroup->>+FilterRule: Predicate rules
    Note over FilterRule: Loop for all rules
    FilterRule->>FilterRule: Rule 1 apply
    FilterRule->>FilterRule: Rule 2 apply
    FilterRule->>FilterRule: Rule 3 apply
    Note over FilterRule: End of loop
    FilterRule->>-FilterGroup: Allowed or denied
    alt FilterGroup allowed or denied
        FilterGroup->>-Consumer: Allowed or denied
    else FilterGroup throws exception
        FilterGroup->>Consumer: Apply default strategy
    end
```
And provide optional default policy to handle filtering exceptions for allowing or denying access.
I have initially designed the following specific types of rules:
[...]

Issue #825, opened 2022-04-06
[Feature] Event Filter


AhahaGe:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/eventmesh/issues?q=is%3Aissue) and found no similar issues.

### Feature Request

1. Scenario
Traditional, when consumer want to filter interested messages,  consumer need to rely on specified MQ client and code upon it to accomplish. But when it comes to apache EventMesh, users only need to **submit customized expressions which defines message filter logic to EventMesh**, EventMesh will do the filter and return filtered messages to users.

2. Implementation

2.1 Implementation One
Do stream processing in EventMesh, it's MQ independent, but have some impact in performance. And we can maintain the expression engine ourself.
- concrete implementation, define a set of EventMesh api to integrate and implement filter rule engine, combine with EQL and Json path. And we can also consider integrating the research rules engine (DataSonnet).

2.2 Bind to RocketMq, use RocketMq as engine to realize filter.
- concrete implementation, implement openmessaging-api interface, and use RocketMq to realize filter ability.

- Conclusion: At present, it is inclined to plan 2.1. The ability of filter has nothing to do with specific MQ components. The specific plan still needs to be investigated.

- related problems:
1. Can Kafka originally support filter?
2. Does RocketMq's filter support multiple thread, currently support.
3. Consider multi-tenant segregation.
4. The storage problem of rules supports plug-in implementation.

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

ruanwenjun:
Plan 2.1 look good to me. I think the performance will not be impacted a lot unless we need to do some join operation.

xwm1992:
Do we need to integrate the rules engine, such as Drools or EasyRule ?

AhahaGe:
> Do we need to integrate the rules engine, such as Drools or EasyRule ?

Sounds good. We can discuss later and I add it into problem list.

jinrongluo:
I have a quick look at Drools and EasyRule, their rules are not for filtering.  I think we need the simple rules like RocketMQ filters

https://rocketmq.apache.org/docs/filter-by-sql92-example/

for example:  $.subject == 'myservice'

We can implement our own simple Rule engine to process this rules using Parse Tree (https://en.wikipedia.org/wiki/Parse_tree) and JsonPath SDK

Concrete Implementation:

1. Parse the Filter Expression into Parse Tree
[...]

PR #4365, opened 2023-08-15
[WIP][ISSUE #4043]initial implementation of filter and transform

pmupkin:
Initial Informal Implementation of Filter and Transform. Link to relevant issues: #825 . #4043 . #3263
## **1. Overview**

Filter and transformer are components of EventMesh used for event filtering and event transformation. During the process of sending and receiving events in EventMesh, rules for event filtering and transformation can be configured to achieve the capability of filtering and transforming events.

Filter refers to the process of filtering events in event subscriptions using configured filtering rules, and then routing the filtered events to event destinations.

Transformer refers to the process of transforming the content of events by configuring types in event subscriptions, converting CloudEvents standard events into event types that event destinations can accept.

Both of these functionalities are implemented using Jackson and JsonPath.

## **Filter**

Filter is an event filtering module. Its primary function is to match events generated by the event source with filtering rules. Only when a successful match occurs, the event will be routed to the event destination associated with the filtering rule. The filtering rule must have the same structure as the matched event.

1. Supported filtering conditions:
    - **Prefix Match**: Specifies matching based on the value of a particular field with a prefix. The keyword is **`prefix`**.
    - **Anything-But Match**: Specifies matching a string-type field with any value except the ones provided. The keyword is **`anything-but`**.
    - **Suffix Match**: Specifies matching based on the value of a particular field with a suffix. The keyword is **`suffix`**.
    - **Numeric Match**: Specifies matching within a range of numeric values. You need to specify an operator and a number, with operators including ">" (greater than), ">=" (greater than or equal to), "<=" (less than or equal to), "<" (less than), "=" (equal to), and "!=" (not equal to). Numeric values are treated as Doubles by default and should be stored in an array format. The keyword is **`numeric`**.
    - **Existence Match**: Checks whether a field exists, with options to set it as true if it exists or false if it doesn't. The keyword is **`exists`**.
    - **Specified Match**: Determines whether a value exists in a particular field. In this type of matching condition, there's no need to specify a value.
    - **Composite Match**: In addition to individual matching rules, Filter also supports combining multiple rules to create more complex event patterns.
    
    Matching between different fields is performed using AND logic, while matching within the same field is performed using OR logic by default.

2. Processing Flow

The implementation of the filter primarily relies on patterns and conditions. The execution flow of the filter is illustrated in the diagram below:
未命名文件 (11)

PatternBuilder: Responsible for parsing the condition content and generating patterns.

Pattern: Evaluates events based on filtering conditions and returns results.

ConditionBuilder: Constructs specific conditions based on different condition keywords.

Condition: Includes various matching conditions, each with its own matching implementation.

3. Usage Example
```java
String condition = "{
    "source":[
        {
            "prefix":"eventmesh."
        },
				{
            "suffix":"eventmesh."
        },
    ]
}";
String event = "{
  "id": "7bf73129-1428-4cd3-a780-95db273d1602",
  "source": "eventmesh.source",
  "time": "2023-11-11T21:29:54Z",
  "data":{
        "name":"filter-test",
        "example":"demo"
    }
}";
Pattern pattern =  PatternBuilder.build(condition);
Boolean res = pattern.filter(cont);
```

## **Transformer**
The transformer is an event conversion module designed to transform CloudEvents standard events into types that the event target can process.

1. Supported Types

Transformer supports three types of transformations:

Direct Transfer: No event transformation is performed; the CloudEvents standard event is directly routed to the target.

Constant: Regardless of the event content, the event triggers the event target without sending the event content to the target.

Variable: Extract variable values from the CloudEvents standard event and route them to the event target according to the format defined in the template. Users can define custom templates and specify the list of variables to extract. The transformer will extract the variables from the event and route them to the event target according to the template's defined format.

2. Processing Flow

As an example of template matching in the transformer, the implementation is as follows:
未命名文件 (12)

3. Usage Example
```java
//ORIGIN
Transformer transformer =new TransformerBuilder.Builder(TransformerType.ORIGINAL).build();
String output = transformer.transform(EVENT);
//Constant
Transformer transformer = new TransformerBuilder.Builder(TransformerType.CONSTANT).setContent("constant test").build();
String output = transformer.transform(EVENT);

```


mxsm (commented):
@pmupkin  Please link this pr to related issue

Alonexc:
image
please check code style.

pmupkin:
> image please check code style.
 Sure, I have fixed the code style issues.

pmupkin:
> @pmupkin Please link this pr to related issue

Okay, I've added some explanations and clarifications, and I've also re-linked the related issues.

xwm1992 (approved):

lrhkobe (approved):
LGTM

PR #4438, opened 2023-09-07
[ISSUE #4396] Realize the admin function for Kafka stroage-plugin

### LINKED ISSUE #4396: [Feature] Realize the admin function for Kafka stroage-plugin
### Search before asking

### Feature Request

image

### Code of Conduct

Pil0tXia:
Fixes #4396.

### Motivation

image

### Modifications

- Supports configuring the number of partitions and ISR replicas to which the topic belongs.
- Has detailed error messages.
- Although `storage.admin` will be migrated to the `admin` module in the near future, this PR is already halfway done, so I finished it up.

image

### Documentation

### DIFF
--- a/eventmesh-storage-plugin/eventmesh-storage-kafka/src/main/java/org/apache/eventmesh/storage/kafka/admin/KafkaAdmin.java
+++ b/eventmesh-storage-plugin/eventmesh-storage-kafka/src/main/java/org/apache/eventmesh/storage/kafka/admin/KafkaAdmin.java
@@ -25,14 +25,15 @@ import org.apache.eventmesh.storage.kafka.config.ClientConfiguration;
 import org.apache.kafka.clients.admin.Admin;
 import org.apache.kafka.clients.admin.DescribeTopicsResult;
 import org.apache.kafka.clients.admin.ListOffsetsResult.ListOffsetsResultInfo;
+import org.apache.kafka.clients.admin.NewTopic;
 import org.apache.kafka.clients.admin.OffsetSpec;
 import org.apache.kafka.clients.admin.TopicDescription;
 import org.apache.kafka.clients.consumer.ConsumerConfig;
 import org.apache.kafka.common.TopicPartition;
 import org.apache.kafka.common.TopicPartitionInfo;
 
-import java.time.Instant;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
@@ -50,8 +51,6 @@ import lombok.extern.slf4j.Slf4j;
 public class KafkaAdmin extends AbstractAdmin {
 
     private static final Properties properties = new Properties();
-
-    private static final long startUpTime = Instant.now().toEpochMilli();
 
     public KafkaAdmin() {
         super(new AtomicBoolean(false));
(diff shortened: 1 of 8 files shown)


eventmesh-storage-plugin/eventmesh-storage-kafka/src/main/resources/kafka-client.properties
```diff
@@ -14,13 +14,21 @@
...
-#######################kafka-client##################
-eventMesh.server.kafka.namesrvAddr=127.0.0.1:9092;127.0.0.1:9092
+
+############################# EventMesh Kafka Client #############################
+
+eventMesh.server.kafka.namesrvAddr=localhost:9092;localhost:9092
 eventMesh.server.kafka.cluster=DefaultCluster
 eventMesh.server.kafka.accessKey=********
 eventMesh.server.kafka.secretKey=********
+# You may see kafka.server.KafkaConfig for additional details and default values
```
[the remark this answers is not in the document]
Pil0tXia: image

This comment is written by someone else and it comes from Kafka: 
https://github.com/apache/kafka/blob/54ab5b29e47fd3f399e0575df2f490a33f12804b/config/server.properties#L17-L18

I think the reason he wrote it that way is that this config file comes from kafka, and there are a lot of configuration items that are specific to kafka, not eventmesh, so you need to refer to the handling of configurations in kafka to understand the configurations further.

I have moved this comment to a more suitable place.

eventmesh-storage-plugin/eventmesh-storage-kafka/src/main/java/org/apache/eventmesh/storage/kafka/admin/KafkaAdmin.java
```diff
@@ -0,0 +1,148 @@
...
+        ClientConfiguration clientConfiguration = configService.buildConfigInstance(ClientConfiguration.class);
+
+        kafkaProps.put(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, clientConfiguration.getNamesrvAddr());
+        topicProps.put("partitionNum", clientConfiguration.getPartitions());
+        topicProps.put("replicationFactorNum", (int) clientConfiguration.getReplicationFactors());
+    }
+
+    @Override
+    public List<TopicProperties> getTopic() {
+        try (Admin client = Admin.create(kafkaProps)) {
```
fabian4, line 74: Can we create the client at the start once for all, instead of we do it each time calling the api.

Pil0tXia: If a shared client is used, then, once an exception occurs in one operation, blocking of this shared client will cause other operations to be blocked as well. After a client is closed, restarting the server is the only way to recreate a usable client, making it impossible to manage its lifecycle.

Pil0tXia:
Made a minor performance optimization by supporting topic batch listing.

eventmesh-storage-plugin/eventmesh-storage-kafka/src/main/java/org/apache/eventmesh/storage/kafka/admin/KafkaAdmin.java
fabian4: Will the client handle it itself just like the producer and consumer cause we don't need to create a new producer each time we send the message. Or just admin client will behavior like this?

Pil0tXia: try-with-resource is the recommended usage of Kafka Admin API on the Kafka official documentation:

https://github.com/apache/kafka/blob/b72d92919f71f7360cd712741c0470759fc10239/clients/src/main/java/org/apache/kafka/clients/admin/Admin.java#L75-L93

You may check out `org.apache.eventmesh.storage.rocketmq.admin.RocketMQAdmin#createMQAdminExt` to find a similar treatment.

xwm1992 (approved):

PR #4486, opened 2023-10-14
 [ISSUE #3520]Swap these 2 arguments so they are in the correct order: expected value, actual value.[EventMeshServerTest]

### LINKED ISSUE #3520: [Enhancement] Swap these 2 arguments so they are in the correct order: expected value, actual value.[EventMeshServerTest]
### Search before asking

### Enhancement Request

image
Located at:
eventmesh-runtime/src/test/java/org/apache/eventmesh/runtime/boot/EventMeshServerTest.java line 63-87
analysis and explanation:
The standard assertions library methods such as org.junit.Assert.assertEquals, and org.junit.Assert.assertSame expect the first argument to be the expected value and the second argument to be the actual value. For AssertJ, it’s the other way around, the argument of org.assertj.core.api.Assertions.assertThat is the actual value, and the subsequent calls contain the expected values. Swap them, and your test will still have the same outcome (succeed/fail when it should) but the error messages will be confusing.

Swap actual and expected values.

HeeHunLee:
fix #3520

### DIFF
--- a/eventmesh-runtime/src/test/java/org/apache/eventmesh/runtime/boot/EventMeshServerTest.java
+++ b/eventmesh-runtime/src/test/java/org/apache/eventmesh/runtime/boot/EventMeshServerTest.java
@@ -60,31 +60,31 @@ public class EventMeshServerTest {
     }
 
     private void assertTCPConfig(EventMeshTCPConfiguration config) {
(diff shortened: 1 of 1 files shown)


mxsm (changes requested):
@HeeHunLee  Please fix code style
image

mxsm (approved):

PR #4539, opened 2023-11-04
try to fix ci error.

pandaapo:
try to fix ci error.


xwm1992:
@pandaapo I have tested this method, change the matrix language can't fix the error.

xwm1992:
> > @pandaapo I have tested this method, change the matrix language can't fix the error.
> 
> Because CodeQL [does not support Rust language](https://docs.github.com/en/code-security/code-scanning/creating-an-advanced-setup-for-code-scanning/customizing-your-advanced-setup-for-code-scanning#changing-the-languages-that-are-analyzed), so I do this try.

But the ci error was caused after the c sdk merged.

PR #4583, opened 2023-11-27
[ISSUE #4581]  http demo rely on tcp runtime

### LINKED ISSUE #4581: [Question] When executing the code in examples, is it necessary to enable the runtime of the tcp protocol?
### Search before asking

### Question

When I run http.demo, if the runtime only supports the HTTP protocol, the subscribe process will not start.

Actually I found that this is a problem caused by the spring sample code. Because the startup of `SpringSourceConnector` relies on the TCP protocol as runtime. 

If I understand wrong please give me some hints.

hhuang1231:
Fixes #4581.

### Motivation

#4581

### Modifications

avoid http demo and grpc demo rely on tcp runtime 

### Documentation


mxsm (approved):
LGTM

PR #4636, opened 2023-12-11
Modified the catch blocks for handling InterruptedExceptions in EtcdMetaService.

### LINKED ISSUE #4096: [Enhancement] InterruptedExceptions should never be ignored in the code.[EtcdRegistryService]
### Search before asking

### Enhancement Request

image
image
located at:
eventmesh-registry-plugin/eventmesh-registry-etcd/src/main/java/org/apache/eventmesh/registry/etcd/service/EtcdRegistryService.java line 165,265
analysis and explanation:

restore interrupted state
image
image

HarshSawarkar:
Fixes #4096 .

### Motivation

Earlier InterruptedExceptions were being being ignored as it were only being logged.

### Modifications

Rewrite the catch blocks by restoring interrupted states through the invocation of Thread.interrupt().

### Documentation


Pil0tXia (approved):
LGTM

harshithasudhakar (approved):

PR #4637, opened 2023-12-11
[ISSUE #4646] Add condition for transforming non-data values

### LINKED ISSUE #4646: [Enhancement] Adding a constraint in the transformer.
### Search before asking

### Enhancement Request

From a design perspective, the transformer needs to avoid converting non-data fields that users may attempt to transform. The original code allowed users to transform fields in the EventMesh; 

Restrictions will be added. If users utilize non-data fields, an exception will be thrown.

### Code of Conduct

pmupkin:
Fixes #4646

From a design perspective, the transformer needs to avoid converting non-data fields that users may attempt to transform. The original code allowed users to transform fields in the EventMesh; restrictions have been added. If users utilize non-data fields, an exception will be thrown.


pmupkin:
> https://github.com/apache/eventmesh/blob/fdd87969e34e6e831a420aeb9cfbc6776ecaeec7/eventmesh-transformer/src/main/java/org/apache/eventmesh/transformer/JsonPathParser.java#L43-L61
> 
> 
> I didn't understand the logic here, could you explain it? The code in line 44 uses Jackson to convert JSON format strings into JsonNode, which seems to have nothing to do with JSON path expressions.

Originally, the tools for serializing JSON (Jackson) and extracting JSON (JsonPath) were both placed within a single class. Looking at it now, the naming indeed can be confusing. It might be beneficial to reorganize them separately in the future.

pmupkin:
> > The `match` function, on the other hand, is used to parse CloudEvents messages using JSONPath.
> 
> https://github.com/apache/eventmesh/blob/fdd87969e34e6e831a420aeb9cfbc6776ecaeec7/eventmesh-transformer/src/main/java/org/apache/eventmesh/transformer/JsonPathParser.java#L74-L83
> 
> 
> @pmupkin
> This is the logic of `match`. The code in line 82 , it retrieves the content specified in the first parameter JSON based on the second parameter, JSON path. The second parameter, JSON path, is obtained from `variablesList` property of `JsonPathParser`. You also said that this attribute stores JSON keys and values without a JSON path, does this mean that the `match` method is invalid?
> 这是`match`的逻辑，在第82行代码，是根据第二个参数即JSON path，获取第一个参数JSON中指定的内容。第二个参数JSON path正是从`JsonPathParser`的`variablesList`属性中获取。您也说该属性保存的是JSON键和值，没有JSON path，所以是不是意味着`match`方法失效了？

Regarding the code at line 82, for instance, if a value in the provided JSON format by the user is $.data.time, the JsonPath tool will be used to match corresponding values in CloudEvents. Could you provide an example illustrating the issue you mentioned?

pmupkin:
> $.data.time



> > if a value in the provided JSON format by the user is $.data.time, ... Could you provide an example illustrating the issue you mentioned?
> 
> @pmupkin I'm a bit confused. Shouldn't the JSON format be like the following: `{"a": "1", "b": {"c": "2"}, "d": []}`? How is `$.data.time` in JSON format? Isn't it a JSON Path expression? Do you mean there is a value like `$.data.time` in JSON?
> 
> 我有些困惑。JSON格式不应该是像这样吗`{"a": "1", "b": {"c": "2"}, "d": []}`？`$.data.time`这怎么是JSON格式呢，这不是JSON Path表达式吗？难道您的意思是在JSON中存在`$.data.time`这样的value？

Here's a fact: $.data.time is stored as a string value,like "{"time":"$.data.time"}" and ultimately, the $.data.time extracted can be used for JsonPath matching. That's my understanding.

Pil0tXia:
May I ask that is this PR ready for review?

PR #4699, opened 2023-12-30
Deprecate `LogUtils` for removal

ppkarwasz:
### Motivation

The `LogUtils` class is not useful because:

 * it wraps every logger call in a `if (logger.isDebugEnabled())` or similar guard. These guards are not necessary since the arguments to the logger calls are simple variables and not computationally complex expressions. They don't provide any performance advantage,
 * if we remove the guards every `LogUtils` method is just a static version of a corresponding `Logger` method,
 * the current implementation it breaks location detection of the logger calls, which will always show `LogUtils` as origin; this problem can be fixed, but it is much simpler to inline all `LogUtils` methods,
 * it prevents the project from using source code rewrite tools like [`rewrite-logging-frameworks`](https://github.com/openrewrite/rewrite-logging-frameworks), which work on the original `Logger` class, not on wrappers.

### Modifications

This PR contains 3 commits:

 1. the first commit deprecates the `LogUtils` class and add Error Prone's [`@InlineMe`](https://errorprone.info/docs/inlineme) annotation to help current class users to rewrite their code automatically,
 2. the second commit is **only** for reviewer's sake: it introduces a simple `errorPronePatch` and `errorPronePatchTest` Gradle task to perform automatic inlining of all `LogUtils` call sites (cf. [Error Prone patching](https://errorprone.info/docs/patching)),
 3. the third commit contains **only automatic changes** performed by ErrorProne.

### Reviewer kit

I would suggest reviewer to:

- check the changes in the `LogUtils` class,
- checkout the code locally and **roll back** one commit,
- using JDK 11 run:
  ```
  ./gradlew clean errorPronePatch
  ./gradlew clean errorPronePatch
  ./gradlew clean :eventmesh-sdks:eventmesh-sdk-java:errorPronePatchTest
  ./gradlew clean :eventmesh-runtime:errorPronePatchTest
  ```
  in order to allow Error Prone to rewrite the codebase,
- after this only one Java source file should contain the string "LogUtils",
- verify that the contents of the working directory are **identical** to the head of this PR, i.e. that the last commit does not introduce **any** changes beyond those performed by Error Prone. 

### Documentation

This PR does not require changes in the documentation. The deprecation of `LogUtils` is documented in Javadoc.

On the other hand this change requires a minor version bump for EventMesh (cf. [semantic versioning pt. 7](https://semver.org/#spec-item-7)).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -95,6 +95,10 @@ allprojects {
                   .exclude('**/org/apache/eventmesh/connector/jdbc/antlr**')
                   .dependsOn(spotlessApply)
 
+    configurations {
+        errorprone.extendsFrom(annotationProcessor)
+    }
+
     dependencies {
         repositories {
             mavenLocal()
@@ -104,6 +108,7 @@ allprojects {
             }
         }
         testImplementation "org.junit.jupiter:junit-jupiter:5.6.0"
+        errorprone "com.google.errorprone:error_prone_core:2.24.0"
     }
 
     spotless {
@@ -138,6 +143,7 @@ allprojects {
     test {
         useJUnitPlatform()
     }
+
 }
 
 task tar(type: Tar) {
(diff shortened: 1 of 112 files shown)


ppkarwasz:
I am closing this PR, since it was integrated into #4698.

Pil0tXia:
@ppkarwasz I merged your branch instead of adding code manully, so the squashed commit of PR will have 2 authors. Thank you very much for your task script!

PR #4719, opened 2024-01-04
[ISSUE #4700] Remove logging backends from runtime deps

### LINKED ISSUE #4700: [Bug] Maven artifacts should not depend on logging backend
### Search before asking

### Environment

Other

### EventMesh version

master

### What happened

Multiple EventMesh artifacts have a runtime dependency on `log4j-core` (a Log4j API implementation) and `log4j-slf4j-impl` (an SLF4J implementation), for example [`eventmesh-common`](https://mvnrepository.com/artifact/org.apache.eventmesh/eventmesh-common/1.10.0-release). These can cause problems to applications that depend on EventMesh artifacts, but use a different logging backend.

Unlike logging **APIs** (e.g. `slf4j-api`, `log4j-api`, `commons-logging`, etc.) that can coexist within the same application, logging **backends** and **bridges** (e.g. `logback-classic`, `log4j-core`, `log4j-to-slf4j`, `log4j-slf4j-impl`, etc.) are mutually exclusive. A couple of examples:

* if an application has as **runtime** dependencies `logback-classic` and `log4j-slf4j-impl` (both are implementation of SLF4J) only one of them will work,
* if an application has as **runtime** dependencies `log4j-to-slf4j` (a bridge from Log4j API to SLF4J) and `log4j-slf4j-impl` (a bridge from SLF4J to Log4j API), both of them will cease to work to prevent an infinite loop.

It is easy to find such examples in practice: an application based both on [`zookeper`](https://mvnrepository.com/artifact/org.apache.zookeeper/zookeeper/3.9.1) and EventMesh, will have both `logback-classic` and `log4j-core` on its classpath.

To prevent this kind of conflicts, the following steps should be necessary:

- [ ] all the logging backends and bridges (an incomplete list: `logback-classic`, `log4j-core`, `log4j-to-slf4j`, `log4j-slf4j-impl`, `log4j-to-jul`, `log4j-jul`, `log4j-over-slf4j`) should be removed from the `runtimeClasspath` and moved to the `testRuntimeOnly` configuration,
- [ ] all `log4j2.xml` configuration files should be removed from the published JARs, so that they not interfere with the configuration provided by applications,
- [ ] the `log4j-core`, `log4j-slf4j-impl` and `log4j-jul` artifacts should be added to the TAR distribution, since it is meant to run as a standalone application.

### How to reproduce

To pinpoint the affected artifacts run:

```sh
./gradlew dependencies
```

and check the `runtimeClasspath` configuration.

### Debug logs

_No response_

### Code of Conduct

ppkarwasz:
Fixes #4700.

### Motivation

As explained in #4700 libraries should only depend on logging APIs (`log4j-api`, `slf4j-api`), not logging implementations (`log4j-core`, `logback-classic`).

This allows application developers to add the logging backend of their choice, without having to exclude multiple transitive dependencies.

### Modifications

This PR:

 * Moves logging backends from the `runtimeClasspath` to `testRuntimeOnly` in EventMesh artifacts,
 * Adds exclusion rules to dependencies that force a specific logging backend: RocketMQ and Spring Boot Starter depend on Logback, while Pulsar depends on Log4j Core,
 * Adds a generic exclusion rule for all known logging backends and bridges between APIs,
 * Rewrites the `dist` task to include Log4j Core and bridges from SLF4J and JUL to the Log4j API.

### Documentation


eventmesh-connectors/eventmesh-connector-spring/build.gradle
```diff
@@ -15,22 +15,16 @@
...
 dependencies {
     api project(":eventmesh-openconnect:eventmesh-openconnect-java")
     implementation project(":eventmesh-common")
     implementation project(":eventmesh-sdks:eventmesh-sdk-java")
-    implementation "org.springframework.boot:spring-boot-starter:$spring_boot_version"
-    implementation "org.springframework.boot:spring-boot-starter-validation:$spring_boot_version"
-    implementation "org.springframework:spring-messaging:$spring_version"
+    implementation "org.springframework.boot:spring-boot-starter-validation"
+    implementation "org.springframework:spring-messaging"
+
```
Pil0tXia, line 24: Why don't we exclude `spring-boot-starter-logging` instead of removing `spring-boot-starter` here?

@yanrongzhen Please have a check

tools/dependency-check/check-dependencies.sh
```diff
@@ -34,14 +34,14 @@ self_modules_txt='tools/dependency-check/self-modules.txt'
...
 
 ./gradlew printProjects | grep '.jar' > "$self_modules_txt"
 
-find "$decompress_conf" -name "*.jar" -exec basename {} \; | uniq | sort > "$all_dependencies_txt"
+find "$decompress_conf" -name "*.jar" -exec basename {} \; | sort | uniq > "$all_dependencies_txt"
 
-grep -wvf "$self_modules_txt" "$all_dependencies_txt" | uniq | sort > "$third_party_dependencies_txt"
+grep -wvf "$self_modules_txt" "$all_dependencies_txt" | sort | uniq > "$third_party_dependencies_txt"
 
 # If the check is success it will return 0
```
Pil0tXia, line 46: It seems switching `sort` and `uniq` won't change the result, but speed up a little, right?

tools/dependency-check/known-dependencies.txt
Pil0tXia, line 1: @xwm1992 Do we have to maintain this lengthy text?

build.gradle
```diff
@@ -160,56 +250,60 @@ task zip(type: Zip) {
...
-    }
-    String[] libJars = java.util.Optional.ofNullable(new File("${rootDir}/dist/lib").list()).orElseGet(() -> new String[0])
-    getAllprojects().forEach(subProject -> {
-        var file = new File("${subProject.projectDir}/gradle.properties")
-        if (!file.exists()) {
+tasks.register('installPlugin') {
+    doLast {
+        if (!new File("${rootDir}/dist").exists()) {
             return
         }
```
Pil0tXia, line 257: I don't quite grasp the alterations made on `dist` and `installPlugin` tasks. There have been several changes in grammar, such as why "doLast" was added in this case?

build.gradle
```diff
@@ -572,6 +605,23 @@ subprojects {
...
+            dependency("org.apache.rocketmq:rocketmq-namesrv:$rocketmq_version") {
+                exclude group: 'ch.qos.logback', name: 'logback-classic'
+            }
+            dependency "org.apache.rocketmq:rocketmq-remoting:$rocketmq_version"
+            dependency "org.apache.rocketmq:rocketmq-srvutil:$rocketmq_version"
+            dependency "org.apache.rocketmq:rocketmq-store:$rocketmq_version"
+            dependency("org.apache.rocketmq:rocketmq-tools:$rocketmq_version") {
+                exclude group: 'ch.qos.logback', name: 'logback-classic'
+            }
         }
```
Pil0tXia, line 625: You've added so much deps from RocketMQ here🤣. Most of them are only used in `eventmesh-storage-plugin/eventmesh-storage-rocketmq/build.gradle`. Can we manage these deps excluding there?

eventmesh-connectors/eventmesh-connector-spring/build.gradle
ppkarwasz: The removal of `spring-boot-starter` was not intentional, I'll add it again.

tools/dependency-check/check-dependencies.sh
ppkarwasz: These are old POSIX commands that not necessarily do what you would expect. E.g. what `uniq` does is:

> Filter **adjacent** matching lines from INPUT (or standard input), writing to OUTPUT (or standard output).

(cf. [man uniq](https://man7.org/linux/man-pages/man1/uniq.1.html), emphasis mine).

That is why you need to sort the file before running `uniq` or you'll end up with duplicate entries.

build.gradle
ppkarwasz: I noticed that some of your tasks (like `printProjects`) are executed regardless of what I put on the command line.

I am not a Gradle expert, so I followed the [Task Configuration Avoidance](https://docs.gradle.org/current/userguide/task_configuration_avoidance.html) guide.

From my experiments:

```groovy
task installPlugin() {
    println "Hello world!"
}
```

is executed **always**, whereas:

```groovy
tasks.register('installPlugin') {
    doLast {
        println "Hello world!"
    }
}
```
is executed only if the task is on the command line.

build.gradle
ppkarwasz: Multiple subprojects use RocketMQ:

- `eventmesh-connectors:eventmesh-connector-rocketmq`
- `eventmesh-retry:eventmesh-retry-rocketmq`
- `eventmesh-storage-plugin:eventmesh-storage-rocketmq`

so I moved the exclusion to the dependency management.

AFAIK RocketMQ does not have a `rocketmq-bom` artifact like ([`junit-bom`](https://mvnrepository.com/artifact/org.junit/junit-bom/5.10.1) or [`spring-framework-bom`](https://mvnrepository.com/artifact/org.springframework/spring-framework-bom/6.1.2)) that could be used in the `imports` section above. Adding [`rocketmq-all`](https://mvnrepository.com/artifact/org.apache.rocketmq/rocketmq-all/5.1.4) to the imports would set the version for other unrelated dependencies.

ppkarwasz:
Something is wrong with Java 11 tests. I am switching this to a "draft" to work out the issues.

Pil0tXia:
> Something is wrong with Java 11 tests. I am switching this to a "draft" to work out the issues.

It seems to be a problem with the third party platform. Not your code.

build.gradle
Pil0tXia: It seems `tasks.register('installPlugin')` equals to `task installPlugin()` and is a newer writting.

I'm afraid that `doLast` will change the original execution sequence. I have to test `dist` and `installPlugins` task to make sure they keep the same behavior with the original one.

build.gradle
Pil0tXia: I think managing RocketMQ deps in these 3 modules will be better. RocketMQ is one of the connectors/storage-plugins supported by EventMesh, but EventMesh Runtime is not implemented by RocketMQ. 

For example, EventMesh supports Pulsar or Kafka too, however you won't see their deps in `eventmesh/build.gradle`.

ppkarwasz: Ok, I'll move those deps back to the submodules.

This is mostly a question of taste, since (at least in Maven) dependency management is not inherited by consumers of the artifacts (only by child POMs).

BTW: I submitted apache/rocketmq#7732 to add a `rocketmq-bom` artifact, which will allow to specify the version of all RocketMQ artifacts at once.

build.gradle
ppkarwasz: @Pil0tXia,

I compared your release 1.10.0 with the one generated by this script and they only differ regarding their logging dependencies (as expected).

Pil0tXia: Good job!

"Task configuration avoidance API" is an updated writting and has backward compatibility. IDEA is also warning the outdated writting, so the new writting should be adopted.

image

---

`doLast()` is also a newer writting, and should be adopted. However, the `printProjects` task is meant to print all modules before each task executes, so I think it doesn't need to apply `doLast()`.

So maybe like this:

image

[What is doLast() for? - Help/Discuss - Gradle Forums](https://discuss.gradle.org/t/what-is-dolast-for/27731/3)

eventmesh-common/build.gradle
```diff
@@ -65,9 +61,8 @@ dependencies {
     testImplementation "com.google.guava:guava"
 
     testImplementation "org.slf4j:slf4j-api"
-    testImplementation "org.apache.logging.log4j:log4j-api"
-    testImplementation "org.apache.logging.log4j:log4j-core"
-    testImplementation "org.apache.logging.log4j:log4j-slf4j2-impl"
+    testRuntimeOnly "org.apache.logging.log4j:log4j-core"
+    testRuntimeOnly "org.apache.logging.log4j:log4j-slf4j2-impl"
```
Pil0tXia, line 65: EventMesh runs as a server most of the time, and we have a `start.sh` script that starts the server by executing the
`org.apache.eventmesh.starter.StartUp` to start the server. Only artifacts under the `eventmesh-sdk` module and the `eventmesh-common` module will be referenced by third-party applications.

When I run the `StartUp` class, I get an error that I think is related to `testRuntimeOnly`.

image

[EventMesh Runtime Quick Start | Apache EventMesh](https://eventmesh.apache.org/docs/instruction/runtime/)

build.gradle
```diff
@@ -139,6 +164,72 @@ allprojects {
...
+configurations {
+    distOnly
+}
+
+dependencies {
+    distOnly "org.apache.logging.log4j:log4j-core:${log4jVersion}"
+    distOnly "org.apache.logging.log4j:log4j-slf4j2-impl:${log4jVersion}"
+}
+
+tasks.register('dist') {
```
Pil0tXia, line 178: I think we should let `dist` task depend on `jar` task, otherwise we will get a `can't load the aclService plugin` error because `eventmesh-security-acl-1.10.0-release.jar` is missing in `plugin/security/acl`.

build.gradle
```diff
@@ -139,6 +164,72 @@ allprojects {
...
+    group = JavaBasePlugin.BUILD_TASK_NAME
+    description "Creates the distribution directory"
```
Pil0tXia, line 180: The value of `JavaBasePlugin.BUILD_TASK_NAME` is `build`, which is a different meaning comparing to the `Creates the distribution directory` description.

I don't think these two lines are particularly necessary, since the task name of the dist task itself already accounts for its meaning.

build.gradle
```diff
@@ -139,6 +164,72 @@ allprojects {
...
+             "eventmesh-security-plugin:eventmesh-security-api",
+             "eventmesh-spi",
+             "eventmesh-starter",
+             "eventmesh-storage-plugin:eventmesh-storage-api",
+             "eventmesh-trace-plugin:eventmesh-trace-api",
+             "eventmesh-webhook:eventmesh-webhook-api",
+             "eventmesh-webhook:eventmesh-webhook-admin",
+             "eventmesh-webhook:eventmesh-webhook-receive"]
+    includedProjects.each { dependsOn("${it}:jar") }
+    outputs.dirs('dist/apps', 'dist/bin', 'dist/conf', 'dist/lib', 'dist/licenses')
```
Pil0tXia, line 197: May I ask what these two lines are for? How did you come up with the idea of creating these two lines, did the IDE prompt you?

I didn't understand the usage of `dependsOn("${it}:jar")`.

It seems `outputs.dirs` equals to `mkdirs()`, but I haven't found much documents and examples proving that.

(A little too cautious with old but working scripts

eventmesh-common/build.gradle
ppkarwasz: Starting EventMesh with `start.sh` should work the same way, since I added the logging backend in the `dist` task.

Now I also restored it in the `eventmesh-starter` artifact, so it should be fine.

build.gradle
ppkarwasz: The first line should inform Gradle that the `:dist` tasks depends on all the `jar` tasks of the projects in includes `:eventmesh-common:jar`, `:eventmesh-meta:eventmesh-meta-api:jar`, etc.

The second line should inform Gradle that the `:dist` task creates and fills those directories. It should help Gradle in determining if `:dist` needs to run or is up-to-date (cf. [Benefits of declaring task inputs and outputs](https://docs.gradle.org/current/userguide/incremental_build.html#sec:task_input_output_side_effects)). Disclaimer: I mostly use Maven in my day job, so my Gradle knowledge is somehow limited.

build.gradle
ppkarwasz: Removed.

build.gradle
```diff
@@ -94,6 +94,33 @@ allprojects {
                   .exclude('**/org/apache/eventmesh/connector/openfunction/client/CallbackServiceGrpc**')
                   .exclude('**/org/apache/eventmesh/connector/jdbc/antlr**')
 
+    configurations {
+        if (!"eventmesh-starter".equals(project.name)) {
+            runtimeClasspath {
+                // Log4j API implementations
+                exclude group: 'org.apache.logging.log4j', module: 'log4j-core'
```
Pil0tXia, line 101: There are actually a lot of executable servers under `eventmesh-connectors` besides `eventmesh-starter` that will be distributed as docker images in the future, and they run into the same problem.

image

Once we add more runnable services in the future, we will need to maintain a list here, with some inflexibility. Therefore, maybe excluding the `eventmesh-sdks` module here is enough, don't you think?

eventmesh-starter/build.gradle
```diff
@@ -17,4 +17,7 @@
 
 dependencies {
     implementation project(":eventmesh-runtime")
-}
\ No newline at end of file
+    // Logging backend
+    runtimeOnly "org.apache.logging.log4j:log4j-core"
+    runtimeOnly "org.apache.logging.log4j:log4j-slf4j2-impl"
```
Pil0tXia, line 22: Adding specific dependencies here is a bit inconsistent with the design here: https://eventmesh.apache.org/docs/instruction/runtime/#341-install-plugins

Since logging backend is shared within project, I hope that it can be added only once in `eventmesh-common` module~.

I searched for "log4j-core" within project, and it shows that some modules under `eventmesh-openconnect` is depending on it directly, which can be substituted with `eventmesh-common` denpendency.

eventmesh-connectors/eventmesh-connector-rocketmq/build.gradle
```diff
@@ -16,21 +16,25 @@
...
+        "org.apache.rocketmq:rocketmq-srvutil:${rocketmq_version}",
+        "org.apache.rocketmq:rocketmq-filter:${rocketmq_version}",
+        "org.apache.rocketmq:rocketmq-acl:${rocketmq_version}",
+        "org.apache.rocketmq:rocketmq-srvutil:${rocketmq_version}"
 ]
 
+configurations {
+    implementation.exclude group: "ch.qos.logback", module: "logback-classic"
+    implementation.exclude group: "ch.qos.logback", module: "logback-core"
+}
```
Pil0tXia, line 36: I'm a little confused as to why some modules have removed the exclusion statements for the Logging backend and others have added them.

Is it better to exclude them uniformly in the root directory, or to exclude them where the dependencies are introduced?

build.gradle
```diff
@@ -139,6 +166,72 @@ allprojects {
...
+def log4jVersion = '2.22.1'
+
+configurations {
+    distOnly
+}
+
+dependencies {
+    distOnly "org.apache.logging.log4j:log4j-core:${log4jVersion}"
+    distOnly "org.apache.logging.log4j:log4j-slf4j2-impl:${log4jVersion}"
+}
```
Pil0tXia, line 178: What is `distOnly` configurations used for here? I didn't find any document related to `distOnly`. It seems to be "Adding logging backend jar to `dist` task only". Removing `configurations.distOnly` seems to have no change.

In the project's root dependency, adding specific dependencies seems to contradict `dependencyManagement`. The approach you've provided currently is more of a subtraction followed by addition, which means excluding all logging backends from the project's root directory first, then adding them back to the `dist` and executable classes. As far as I know, most projects adopt a more traditional approach, which is addition followed by subtraction, meaning first introducing logging backends normally, then excluding them from the SDK.

The former approach of "adding logging backends to `dist` and executable classes" overly emphasizes the presence of a specific logging backend and disperses the dependencies of logging backends in multiple places, including `eventmesh-starter`, `eventmesh-connectors`, `dist`, etc. The latter, in the step of "introducing logging backends normally", only requires adding the logging backend the project is using once in `eventmesh-common`.

Personally, I tend to be more conservative and might prefer the latter approach. However, if the former could have a more understandable solution (especially for newcomers to the community), I would also choose the former.

build.gradle
Pil0tXia: Thank you! This explanation is very detailed.

>The first line should inform Gradle that the `:dist` tasks depends on all the `jar` tasks of the projects in includes `:eventmesh-common:jar`, `:eventmesh-meta:eventmesh-meta-api:jar`, etc.

It seems that this line didn't work, according to https://github.com/apache/eventmesh/pull/4719#discussion_r1554553497.

>The second line should inform Gradle that the `:dist` task creates and fills those directories.

For the `dist` task, I don't think we need to declare its output folder. Because when we execute the `dist` task, we definitely want it to repack a copy. Besides, the `dist` task is a terminal task, and there won't be any more tasks depending on it.

build.gradle
ppkarwasz: Nice catch!

Actually `eventmesh-security-acl` is not copied by the `dist` task, but the `installPlugin` task. I fixed the dependencies of the latter in 8d5cab712b496b5166980d9f103075eb8f2c31ca.

build.gradle
ppkarwasz: > > The first line should inform Gradle that the `:dist` tasks depends on all the `jar` tasks of the projects in includes `:eventmesh-common:jar`, `:eventmesh-meta:eventmesh-meta-api:jar`, etc.
> 
> It seems that this line didn't work, according to [#4719 (comment)](https://github.com/apache/eventmesh/pull/4719#discussion_r1554553497).

Fixed in 8d5cab712b496b5166980d9f103075eb8f2c31ca, the problem was the dependencies of the `installPlugin` task.

> > The second line should inform Gradle that the `:dist` task creates and fills those directories.
> 
> For the `dist` task, I don't think we need to declare its output folder. Because when we execute the `dist` task, we definitely want it to repack a copy. Besides, the `dist` task is a terminal task, and there won't be any more tasks depending on it.

Doesn't `installPlugin` and `tar` depend on `dist`? I can replace the `outputs.dirs` with `mkdirs` if you prefer.

eventmesh-connectors/eventmesh-connector-rocketmq/build.gradle
ppkarwasz: There are two ways logging backends appear in EventMesh artifacts:

- as **direct** dependencies in your modules. I have removed all of those, the `configurations { runtimeClasspath {} }` block in the root project should just prevent reintroducing them by mistake,
- as **transitive** dependencies of your **direct** dependencies. For example [`rocketmq-broker`](https://mvnrepository.com/artifact/org.apache.rocketmq/rocketmq-broker/4.9.5) seems to be such a dependency. Since it depends on Logback, all your artifacts that use RocketMQ effectively depend on Logback.

I tested the second point by running:
```sh
./gradlew generatePomFileForMavenJava
```
and compared the generated `eventmesh-storage-plugin/eventmesh-storage-rocketmq/build/publications/mavenJava/pom-default.xml` files. The POM file **before** the introduction of these exclusions contained just:
```xml
  <dependencies>
    ...
    <dependency>
      <groupId>org.apache.rocketmq</groupId>
      <artifactId>rocketmq-broker</artifactId>
      <version>4.9.5</version>
      <scope>runtime</scope>
    </dependency>
    ...
  </dependencies>
```

After the changes it contained exclusions:
```xml
  <dependencies>
    ...
    <dependency>
      <groupId>org.apache.rocketmq</groupId>
      <artifactId>rocketmq-broker</artifactId>
      <version>4.9.5</version>
      <scope>runtime</scope>
      <exclusions>
        <exclusion>
          <artifactId>logback-classic</artifactId>
          <groupId>ch.qos.logback</groupId>
        </exclusion>
        <exclusion>
          <artifactId>logback-core</artifactId>
          <groupId>ch.qos.logback</groupId>
        </exclusion>
      </exclusions>
    </dependency>
    ...
  </dependencies>
```

build.gradle
ppkarwasz: I am not quite sure how to deal with those artifacts.

The `eventmesh-starter` artifact has **only** one possible usage: it must be started from a command line.

On the other hand artifacts like `eventmesh-runtime` or the connectors might be used by users to start EventMesh programmatically, right? In that case I wouldn't impose `log4j-core` as a dependency.

In your Docker images the logging backend can be added additionally to the "normal" dependencies, just like it is added in the `tar.gz` distribution.

If you think external projects should only use `eventmesh-sdk-java` or `eventmesh-openconnect-api`, I could move the exclusions to those projects only. What do you think?

eventmesh-starter/build.gradle
ppkarwasz: Most `log4j-core` dependencies should be in the `testRuntimeOnly` configuration, so they are invisible to external projects that use those `eventmesh-openconnect` artifacts.

build.gradle
Pil0tXia: `installPlugin` is a seperate task (according to https://eventmesh.apache.org/docs/instruction/runtime/#24-package-plugins), which can bring more flexibility in plugin file loading. So I think `installPlugin` doesn't need to depend on `dist`, and `dist` should depend on `jar`.

In terms of task scripts, I think there isn't any logic that needs to be changed, it's better to minimize logic changes ~ the grammar change is totally ok.

eventmesh-starter/build.gradle
Pil0tXia: >it shows that some modules under `eventmesh-openconnect` is depending on it directly, which can be substituted with `eventmesh-common` denpendency.

What I mean is that `eventmesh-openconnect` should have depended on `eventmesh-common` to avoid having to manually add dependencies in `eventmesh-openconnect`. This was an oversight by previous developers. `eventmesh-openconnect` module should be less likely to be depended on by third-party applications.

eventmesh-connectors/eventmesh-connector-rocketmq/build.gradle
Pil0tXia: If `configurations { runtimeClasspath {} }` block in the root project should be able to avoid the transitive logging backend passed by direct dependencies globally, then there's no need to exclude Logback again where the RocketMQ dependencies are introduced.

Some `implementation.exclude` statements are removed (like `eventmesh-connectors/eventmesh-connector-rabbitmq/build.gradle`) and I think that's because `com.rabbitmq:amqp-client` won't introduce any logging backend and these statements are redundant, right?

build.gradle
Pil0tXia: `eventmesh-starter` will start `eventmesh-runtime`, and the connectors can also be started from a command line.

I think external projects should only use `eventmesh-sdk-java` or any other language of SDKs under `eventmesh-sdks` (which isn't written in java). `eventmesh-openconnect-api` is less likely to be used in external projects.

Just as I said in https://github.com/apache/eventmesh/pull/4719#discussion_r1554660695, I may prefer the latter approach (first introducing logging backends normally in `eventmesh-common`, then excluding any transitive logging backend from direct SDK dependencies), because excluding logging backends globally won't resolve all problems, but also introducing more configurations on a single artifact in many scattered places, which is not that elegant. What do you think then? ☺️

I really appreciate your thoughts, but I'm sorry that I didn't explain the various uses of the project to you in the first place, which resulted in the solution you were trying to implement in the first place not being as elegant after the modifications.

ppkarwasz:
@Pil0tXia,

Given your explanations above, I'll refactor this PR so that:

- your artifacts do not have **transitive** dependencies on logging backends. I'll explicitly exclude both Logback and Log4j Core from the external projects that break this rule (RocketMQ and Zookeeper),
- exclude logging backends from the **direct** dependencies of `eventmesh-jdk-java`, so that consumers of that artifact can choose their own logging backend,
- I'll **keep** the logging backends that are **explicitly** declared in your other artifacts.

Does this sound like a plan?

Pil0tXia:
@ppkarwasz 

>Does this sound like a plan?

Yeah!

>your artifacts do not have **transitive** dependencies on logging backends.

My understanding of this statement is that every eventmesh submodule, except `eventmesh-common`, should not pass logging backend dependencies to other modules. Is that correct? I think this is fine.

>I'll **keep** the logging backends that are **explicitly** declared in your other artifacts.

Yeah, except `eventmesh-openconnect` according to https://github.com/apache/eventmesh/pull/4719#discussion_r1554823665. `eventmesh-common` module can pass Slf4j and Log4j2 to other modules.

build.gradle
ppkarwasz: I restored the dependency of `:dist` on all `:*:jar` tasks.

eventmesh-connectors/eventmesh-connector-rocketmq/build.gradle
ppkarwasz: I did some experiments and the `configuration { runtimeClasspath {} }` block might prevent the propagation of logging backends in the EventMesh project, but it is ineffective for third-party projects.

A better alternative would be the [Gradle Enforcer Plugin](https://kordamp.org/enforcer-gradle-plugin/) configured with a [`BannedDependencies`](https://kordamp.org/enforcer-gradle-plugin/#_banneddependencies) rule. Unfortunately I was not able to configure it without triggering the kordamp/enforcer-gradle-plugin#22 issue.

In the end, I removed the `configurations` block entirely and replaced it with surgical strikes against broken dependencies (RocketMQ, Pulsar and Zookeeper). I also excluded `spring-boot-starter-logging` from Spring Boot.

Pil0tXia: I love your use of "surgical strikes"😄

eventmesh-openconnect/eventmesh-openconnect-java/build.gradle
```diff
@@ -32,4 +29,7 @@ dependencies {
 
     compileOnly 'org.projectlombok:lombok'
     annotationProcessor 'org.projectlombok:lombok'
+
+    testRuntimeOnly "org.apache.logging.log4j:log4j-core"
+    testRuntimeOnly "org.apache.logging.log4j:log4j-slf4j2-impl"
```
Pil0tXia, line 34: How about depending on `eventmesh-common` module? To achieve "the logging backends are now only in `eventmesh-common`".

build.gradle
```diff
@@ -139,7 +139,59 @@ allprojects {
     }
 }
 
-task tar(type: Tar) {
+tasks.register('dist') {
+    subprojects.forEach { subProject ->
+        dependsOn("${subProject.path}:jar")
+    }
```
Pil0tXia, line 145: I tried using the basic `dependsOn` statement, but then the plugin jar packages is incomplete.

```suggestion
    dependsOn("jar")
```

The existing writting is more complicated than the old one, is this the simplest writting yet?

build.gradle
```diff
@@ -139,7 +139,59 @@ allprojects {
...
+             "eventmesh-starter",
+             "eventmesh-storage-plugin:eventmesh-storage-api",
+             "eventmesh-trace-plugin:eventmesh-trace-api",
+             "eventmesh-webhook:eventmesh-webhook-api",
+             "eventmesh-webhook:eventmesh-webhook-admin",
+             "eventmesh-webhook:eventmesh-webhook-receive"]
+    outputs.dirs('dist/apps', 'dist/bin', 'dist/conf', 'dist/lib', 'dist/licenses')
+    doLast {
+        includedProjects.each {
+            def project = findProject(it)
```
Pil0tXia, line 164: Better name it `subProject` to indicate that all `from project...` equals to `${projectDir}/...` and all `into 'dist/xxx'` equals to `${rootDir}/dist/xxx`.

---

Maybe `${projectDir}` and `${rootDir}` will be useful to indicate the relative and absolute path, because I spent 5 mins to prove the change equals to the old one. 😁 But the current writting do looks like more concise, so it depends on you.

build.gradle
```diff
@@ -139,7 +139,59 @@ allprojects {
...
+                from project.sourceSets.main.resources.srcDirs
+                into 'dist/conf'
+                duplicatesStrategy = DuplicatesStrategy.EXCLUDE
+                exclude 'META-INF'
+            }
+            copy {
+                from project.configurations.runtimeClasspath
+                into 'dist/lib'
+                exclude 'eventmesh*'
+            }
```
Pil0tXia, line 185: I don't know if it equals to the original writting:

```gradle
            copy {
                into("${projectDir}/dist/lib")
                from project.configurations.runtimeClasspath
            }
            if (rootProject.contains(project.name)) {
                copy {
                    into "${rootDir}/dist/lib"
                    from "${projectDir}/dist/lib"
                    exclude "eventmesh-*"
                }
            }
```

build.gradle
```diff
@@ -160,50 +212,48 @@ task zip(type: Zip) {
     }
 }
 
-task installPlugin() {
-    if (!new File("${rootDir}/dist").exists()) {
-        return
+tasks.register('installPlugin') {
+    // Compute dependency and outputs eagerly
+    dependsOn('dist')
```
Pil0tXia, line 217: I think `installPlugin` doesn't need to depend on `dist` or `jar`, according to https://github.com/apache/eventmesh/pull/4719#discussion_r1554661834.

Depending on `dist` will significantly lengthen the time of task execution. So the original `exists()` check will be better.

build.gradle
```diff
@@ -139,7 +139,59 @@ allprojects {
...
+             "eventmesh-runtime",
+             "eventmesh-security-plugin:eventmesh-security-api",
+             "eventmesh-spi",
+             "eventmesh-starter",
+             "eventmesh-storage-plugin:eventmesh-storage-api",
+             "eventmesh-trace-plugin:eventmesh-trace-api",
+             "eventmesh-webhook:eventmesh-webhook-api",
+             "eventmesh-webhook:eventmesh-webhook-admin",
+             "eventmesh-webhook:eventmesh-webhook-receive"]
+    outputs.dirs('dist/apps', 'dist/bin', 'dist/conf', 'dist/lib', 'dist/licenses')
```
Pil0tXia, line 161: I tried to delete this statement and nothing changes, so maybe there's no need to keep it~

eventmesh-connectors/eventmesh-connector-rocketmq/build.gradle
```diff
@@ -16,25 +16,33 @@
...
 ]
 
 dependencies {
     api project(":eventmesh-openconnect:eventmesh-openconnect-java")
     implementation project(":eventmesh-common")
     implementation rocketmq
+    implementation("org.apache.rocketmq:rocketmq-broker:$rocketmq_version") {
+        // Remove logging backend implementations
+        exclude group: 'ch.qos.logback', module: 'logback-classic'
+    }
```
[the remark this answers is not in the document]
ppkarwasz: I chose this way by observing the generated POM and `*.module` files that are published to Maven Central.

The `configurations` solution you propose gives a POM file like [`eventmesh-storage-rocketmq-1.10.0-release.pom`](https://repo1.maven.org/maven2/org/apache/eventmesh/eventmesh-storage-rocketmq/1.10.0-release/eventmesh-storage-rocketmq-1.10.0-release.pom) and a module file like [`eventmesh-storage-rocketmq-1.10.0-release.module`](https://repo1.maven.org/maven2/org/apache/eventmesh/eventmesh-storage-rocketmq/1.10.0-release/eventmesh-storage-rocketmq-1.10.0-release.module). As you can see, `logback-classic` is excluded from **each** dependency.

By adding the exclusion to `rocketmq-broker`, only `rocketmq-broker` has it, while other dependencies have no exclusions.

**Remark**: the problem with RocketMQ will be solved, when apache/rocketmq#5347 is fixed, although nothing indicates it will ever be solved. The newest version of `rocketmq-broker` replaces the dependency on Logback with a dependency on:

```xml
<dependency>
    <groupId>io.github.aliyunmq</groupId>
    <artifactId>rocketmq-shaded-slf4j-api-bridge</artifactId>
    <version>1.0.0</version>
</dependency>
```

that will need to be excluded as well.

Do you really need `rocketmq-broker` and the other dependencies? Looking at the code, all you need it `rocketmq-tools`.

build.gradle
ppkarwasz: Unless I am mistaken `dependsOn("jar")` means that the `:dist` task from the root project will depend on the `:jar` task from the root project. Here I accentuate that it depends on **all** tasks of type `jar`, e.g. `:eventmesh-common:jar`.

eventmesh-connectors/eventmesh-connector-rocketmq/build.gradle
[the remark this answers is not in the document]
Pil0tXia: >although nothing indicates it will ever be solved

🤣

>As you can see, `logback-classic` is excluded from **each** dependency.

Gradle isn't Maven after all, and while the pom file generated by the `configurations` block isn't accurate from a Maven perspective, it does simplify configuration when using Dependency Lists in Gradle.

So I think we'll just use the `configurations` block and keep the List intact in the build.gradle that uses the `List`. Other build.gradles that don't use the `List` keep the surgical strikes.

>Do you really need `rocketmq-broker` and the other dependencies?

There really is no reference to the `rocketmq-broker` package in the code. I'll create an issue for this, and there will be no need to make changes to this issue in this PR for now.

build.gradle
Pil0tXia: >Here I accentuate that it depends on **all** tasks of type `jar`, e.g. `:eventmesh-common:jar`.

Yeah I understand your code's meaning.

>Unless I am mistaken `dependsOn("jar")` means that the `:dist` task from the root project will depend on the `:jar` task from the root project.

The original `dist` task did depend on the `:jar` task from the root project and it works. I think it should be the same with your way. Is it because the use of the `task.register` writting makes it impossible to depend on the `jar` task from the root project anymore?

eventmesh-connectors/eventmesh-connector-rocketmq/build.gradle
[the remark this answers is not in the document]
ppkarwasz: I restored the list and put a comment with the link to the RocketMQ issue that will allow you to remove the exclusion.

build.gradle
ppkarwasz: No, they are not.

In the original **every** sub project had its own `dist` task that created temporary directories, such as `eventmesh-common/dist/lib`.

Now there is only **one** `dist` task that creates a single distribution in the `dist` folder of the main project.
Were the old `<project_name>/dist/lib` folders useful? I didn't think so, since they didn't contain enough data to make standalone distributions of the sub-projects.

build.gradle
ppkarwasz: In the current build script on `master`, the `dist` task is declared inside `allprojects`, therefore each sub project has its own `dist` task that depends on its own `jar` task.

After the rewrite, there is only one `dist` task. Should I split the `dist` task again among projects?

build.gradle
ppkarwasz: Fixed

build.gradle
ppkarwasz: You are right, the [`Copy`](https://docs.gradle.org/current/dsl/org.gradle.api.tasks.Copy.html) task must create those directories on demand.

I removed the line.

build.gradle
Pil0tXia: >Were the old `<project_name>/dist/lib` folders useful?

Maybe not I guess.

>exclude 'eventmesh*'

It's different with the original `exclude 'eventmesh-*'`. Is this intended?

build.gradle
```diff
@@ -160,50 +211,46 @@ task zip(type: Zip) {
...
+    var pluginProjects = subprojects.findAll {
+        it.file('gradle.properties').exists()
+                && it.properties.containsKey('pluginType')
+                && it.properties.containsKey('pluginName')
+    }
+    pluginProjects.
+            forEach(subProject -> {
+                var pluginType = subProject.properties.get('pluginType')
+                var pluginName = subProject.properties.get('pluginName')
+                dependsOn("${subProject.path}:jar")
```
Pil0tXia, line 224: subProjects have already depended on `jar` in `dist` task. Is it necessary to declare it again in task `installPlugin`?

build.gradle
```diff
@@ -160,50 +211,46 @@ task zip(type: Zip) {
...
+                outputs.dir("dist/plugin/${pluginType}/${pluginName}")
```
Pil0tXia, line 225: The `outputs.dirs` of `dist` task has been removed. Is it useful to keep this one here in task `installPlugin`?

build.gradle
ppkarwasz: It has almost the same effect, but it is rather a typo.

build.gradle
ppkarwasz: Given [your previous comment](https://github.com/apache/eventmesh/pull/4719#discussion_r1557893664) regarding the dependance of `installPlugin` on `dist`, I am a little bit confused about the relationship between these tasks.

Is `installPlugin` always called after `dist`? If so, it should probably depend on it. On the other hand `dist` does not need to depend on `:eventmesh-connectors:eventmesh-connector-rocketmq:jar` for example, because it does not use that JAR.

What happens in the `master` branch is:

- there is a `dist` task for **each** project, even "plugin" projects,
- there is a **single** `:installPlugin` task that collects all the plugins into the `dist` folder.

Maybe it should be reversed:

- there should be a **single** `:dist` project that builds up the basic binary distribution (this is implemented in this PR),
- there should be a `installPlugin` task for **each** "plugin" project (currently in this PR, there is a single one).

This way running:
```sh
./gradlew clean :eventmesh-connectors:eventmesh-connector-rocketmq:installPlugin
```
would create a base distribution with just a single plugin. What do you think?

ppkarwasz: > > If so, it should probably depend on it.
> 
> IMO, since `jar` tasks have already been executed in `dist` task, so they shouldn't be executed again in task `installPlugin`, just as the original `installPlugin` task didn't do that either.

You can not be sure that `dist` has already been executed, since `installPlugin` does not have a dependency on `dist`. These dependencies ensure that those JARs are ready.

If `dist` **has** been already executed. The JARs will not be built again. If you run
```
./gradlew --console plain --no-daemon clean dist installPlugin
```
nothing will be executed between the `:dist` task and `:installPlugin`. If multiple tasks in the build depend on task `foo`, the task will be executed only once (or most probably its result will be retrieved from the cache).

> > would create a base distribution with just a single plugin
> 
> That's a good idea, but I'm not sure if anyone does this though, because trying to remember the exact name of a plugin is difficult, there are a lot of plugins, and the vast majority of api's require a default plugin.
> 
> After you reminded me of this, I realized that the `installPlugin` task is actually doing the work of the `installPlugins` task. However, considering that we don't need an new `installPlugin` task, changes to the task would involve multiple changes to documentation, CI, etc., so I don't think it's really necessary.

I could do this in another PR.

Pil0tXia: >You can not be sure that `dist` has already been executed, since `installPlugin` does not have a dependency on `dist`. These dependencies ensure that those JARs are ready.

Then `installPlugin` should also depend on the `dist` task, not the `jar` task. The `installPlugin` task won't get the full result if the `dist` task hasn't been executed.

However, if the `installPlugin` depends on the `dist` task, the user will not be able to load their own jar file, losing the flexibility of the `installPlugin` task.

Considering that both the `dist` task and the `installPlugin` task will be executed (and should be) anyway, I don't think the `installPlugin` task should depend on the `dist` task.

>If multiple tasks in the build depend on task `foo`, the task will be executed only once (or most probably its result will be retrieved from the cache).

When testing your branch on my own PC, sometimes the `installPlugin` task executes a long `compileJava` task and outputs some deprecated warnings, even though the `dist` task has already been executed. This problem may be worse on less powerful machines. So I feel that the dependency of the `installPlugin` task on the `dist` task introduces some unnecessary performance overhead.

>I could do this in another PR.

I'm don't know how I'm going to introduce this new feature in documentation ...... If I need to load a custom closed-source plugin, maybe I would manually copy the jar package to the specified location in the file explorer after having run the `dist` task, and then run the `installPlugin` task. 

A new task brings some flexibility, however another concern I have is that renaming `installPlugin` to
`installPlugins` and then adding a new `installPlugin` task that is not the same will confuse other developers in the community.

@xwm1992 What do you think?

Pil0tXia: PMCs may be busy these days. How about we just keep `installPlugin`'s original task dependency logic for now, i.e., no dependencies on `dist` or `jar` tasks? Changes to the Gradle task logic can be addressed in another PR.

ppkarwasz: I removed the dependencies.

Pil0tXia (approved):
LGTM. Thanks to @ppkarwasz for his contribution over time.

In summary: This PR mainly modernizes the project's dependency on the logging backends and some obsolete gradle syntax.

- The vast majority of the changes to the Gradle tasks are equivalent to their effects before the modifications, the biggest change being the change of the `dist` task from a submodule task to a root module task.
- For users, the problems mentioned in https://github.com/apache/eventmesh/issues/4700#issuecomment-1997989505 have been solved.

eventmesh-openconnect/eventmesh-openconnect-offsetmgmt-plugin/build.gradle
```diff
@@ -30,4 +27,7 @@ dependencies {
 
     compileOnly 'org.projectlombok:lombok'
     annotationProcessor 'org.projectlombok:lombok'
+
+    testRuntimeOnly "org.apache.logging.log4j:log4j-core"
+    testRuntimeOnly "org.apache.logging.log4j:log4j-slf4j2-impl"
```
Pil0tXia, line 32: here

eventmesh-openconnect/eventmesh-openconnect-offsetmgmt-plugin/eventmesh-openconnect-offsetmgmt-api/build.gradle
```diff
@@ -29,4 +26,7 @@ dependencies {
     testAnnotationProcessor 'org.projectlombok:lombok'
 
     testImplementation "org.mockito:mockito-core"
+
+    testRuntimeOnly "org.apache.logging.log4j:log4j-core"
+    testRuntimeOnly "org.apache.logging.log4j:log4j-slf4j2-impl"
```
Pil0tXia, line 31: and here

Pil0tXia (commented):
Hi~I just found that two submodules are missing the dependency on `eventmesh-common`. It might have been overlooked while resolving https://github.com/apache/eventmesh/pull/4719#discussion_r1557812935. It would be great if you could kindly consider committing it~

eventmesh-openconnect/eventmesh-openconnect-offsetmgmt-plugin/eventmesh-openconnect-offsetmgmt-api/build.gradle
ppkarwasz: Done

Pil0tXia (approved):

eventmesh-examples/build.gradle
```diff
@@ -17,15 +17,14 @@
...
-
 dependencies {
     implementation project(":eventmesh-sdks:eventmesh-sdk-java")
     implementation project(":eventmesh-common")
     implementation project(":eventmesh-storage-plugin:eventmesh-storage-api")
     implementation project(":eventmesh-connectors:eventmesh-connector-spring")
+    implementation('org.springframework.boot:spring-boot-starter') {
+        exclude module: 'spring-boot-starter-logging'
+    }
     implementation 'org.springframework.boot:spring-boot-starter-web'
```
Pil0tXia, line 28: `logback` was introduced back by `spring-boot-starter-web`. I will remove it completely in https://github.com/apache/eventmesh/pull/4831/commits/be06ef7441d58b6cf148ca6fe7457525502fcf4d and https://github.com/apache/eventmesh/pull/4827/commits/0261ef6174d0a431074226b96589c375b4f11242.

PR #4752, opened 2024-01-20
[ISSUE #4095] Code Optimization and Interrupeted Exception handling.[EtcdCustomService]

### LINKED ISSUE #4095: [Enhancement] Code optimization. [EtcdCustomService]

scwlkq:
Fixes #4095 .

### Motivation

Code Optimization and Interrupeted Exception handling in EtcdCustomService.class

### Documentation

### DIFF
--- a/eventmesh-meta/eventmesh-meta-etcd/src/main/java/org/apache/eventmesh/meta/etcd/service/EtcdCustomService.java
+++ b/eventmesh-meta/eventmesh-meta-etcd/src/main/java/org/apache/eventmesh/meta/etcd/service/EtcdCustomService.java
@@ -67,6 +67,9 @@ public class EtcdCustomService extends EtcdMetaService {
                 return eventMeshServicePubTopicInfoList;
             }
         } catch (Exception e) {
+            if(e instanceof InterruptedException) {
+                Thread.currentThread().interrupt();
+            }
             log.error("[EtcdRegistryService][findEventMeshServicePubTopicInfos] error", e);
             throw new MetaException(e.getMessage());
         }
@@ -91,6 +94,9 @@ public class EtcdCustomService extends EtcdMetaService {
                 return eventMeshAppSubTopicInfo;
             }
         } catch (Exception e) {
+            if(e instanceof InterruptedException) {
+                Thread.currentThread().interrupt();
+            }
             log.error("[EtcdRegistryService][findEventMeshAppSubTopicInfoByGroup] error, group: {}", group, e);
             throw new MetaException(e.getMessage());
         }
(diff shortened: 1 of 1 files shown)


Pil0tXia:
Please correct typos in PR title.

Pil0tXia:
Please fix checkstyle warnings. You may read contribution guidelines to get code lints in your IDE.

eventmesh-meta/eventmesh-meta-etcd/src/main/java/org/apache/eventmesh/meta/etcd/service/EtcdCustomService.java
```diff
@@ -67,6 +67,9 @@ public List<EventMeshServicePubTopicInfo> findEventMeshServicePubTopicInfos() th
                 return eventMeshServicePubTopicInfoList;
             }
         } catch (Exception e) {
+            if (e instanceof InterruptedException) {
+                Thread.currentThread().interrupt();
+            }
             log.error("[EtcdRegistryService][findEventMeshServicePubTopicInfos] error", e);
             throw new MetaException(e.getMessage());
         }
```
Pil0tXia, line 75: Please catch `InterruptedException` in a separate `catch` code block instead of judging it. You may see `org.apache.eventmesh.meta.etcd.service.EtcdMetaService` L173-175 and L300-302 for examples.

https://github.com/apache/eventmesh/blob/21e4c3a36816576de7eec903c885cf6e5f956c31/eventmesh-meta/eventmesh-meta-etcd/src/main/java/org/apache/eventmesh/meta/etcd/service/EtcdMetaService.java#L173-L179

https://github.com/apache/eventmesh/blob/21e4c3a36816576de7eec903c885cf6e5f956c31/eventmesh-meta/eventmesh-meta-etcd/src/main/java/org/apache/eventmesh/meta/etcd/service/EtcdMetaService.java#L298-L305

scwlkq: get

Pil0tXia (approved):
LGTM

Pil0tXia (approved):

mxsm (approved):
LGTM

PR #4759, opened 2024-01-28
[ISSUE 1013] Java sdk split multiple sdk by communication protocol

### LINKED ISSUE #1013: [Enhancement] Java sdk split multiple sdk by communication protocol
### Search before asking

### Enhancement Request

Java sdk split multiple sdk by communication protocol,  implement the slimming of the java sdk.

Java sdk split multiple sdk by communication protocol.

scwlkq:
Fixes #1013 

### Motivation

Java sdk split multiple sdk by communication protocol

### Modifications

1. I found that the EventMeshCommon class in the original TCP package was referenced by the classes of grpc, HTTP, and other modules, so I placed this class under the Common module and corrected the reference.This also eliminates the coupling between the modules of the Java SDK.
2. I will split the original SDKs. 
image
However, the Catalog requires the use of the selector and GRPC modules, so they are placed together. The same applies to the WorkFlow module and selector module.
image
image
3. A small question: The build.gradle files in new module copyed from original Java-sdks. If they have duplicate dependencies, i will delete unnecessary dependencies.

*I am not sure if this segmentation method is correct. If there are any shortcomings, please let me kown, and i will correct it.*

### Documentation

- Does this pull request introduce a new feature? ( no)


Pil0tXia:
I think the naming of `eventmesh-sdk-java-tcp` will be better than `eventmesh-tcp-sdk-java`. What do you think?

scwlkq:
> I think the naming of `eventmesh-sdk-java-tcp` will be better than `eventmesh-tcp-sdk-java`. What do you think?

ye，I think your naming method looks better.

Pil0tXia:
>too much sdks for users, doesn't it?

makes sense.

cc @xwm1992

xwm1992:
Is it necessary to split java-sdk into so many modules?

mxsm:
My personal opinion is to divide them according to different languages, such as Rust, Java, and Go. Instead of dividing them for each protocol, dividing them too finely will lead to a very bloated project. At the same time, whether other language SDKs should also be divided in this way.

Pil0tXia:
What is the original intention of issue #1013?

PR #4764, opened 2024-02-01
[ISSUE #4763] Do some code optimization.[EventMeshCommon]

### LINKED ISSUE #4763: [Enhancement] Do some code optimization.[EventMeshCommon]
### Search before asking

### Enhancement Request

As [comment](https://github.com/apache/eventmesh/issues/1013#issuecomment-1920798532):
I found that the EventMeshCommon class in the original TCP package was referenced by the classes of grpc, HTTP, and other modules. So should I placed this class under the Common module and corrected the reference? This also eliminates the coupling between the modules of the Java SDK.

Now reference:
image

Move EventMesh class to common module, and check original references.

### Code of Conduct

scwlkq:
Fixes #4763 

### Motivation

move EventMeshCommon class to Common module.

### Documentation


Pil0tXia (commented):
Not all members in the `EventMeshCommon` class are shared across the entire project. Some members are used in the `client.tcp` package, while others are used in both the sdk module and the `client` package. 

Therefore, it is not a good idea to directly move the `EventMeshCommon` class, which originally belongs to the `client.tcp` package, to the `common` module. It requires a more detailed and clear separation to unify all related usages.

Pil0tXia:
Is there any obstacles in this PR?

scwlkq:
> Is there any obstacles in this PR?

Closed because I thought the issue was meaningless hahaha.

eventmesh-common/src/main/java/org/apache/eventmesh/common/EventMeshCommon.java
```diff
@@ -29,16 +29,5 @@ public class EventMeshCommon {
...
-    public static final String USER_AGENT_PURPOSE_PUB = "pub";
-
-    /**
-     * USERAGENT for SUB
-     */
-    public static final String USER_AGENT_PURPOSE_SUB = "sub";
-
     public static final String EM_MESSAGE_PROTOCOL_NAME = "eventmeshmessage";
-    public static final String OPEN_MESSAGE_PROTOCOL_NAME = "openmessage";
 }
```
Pil0tXia, line 33: `OPEN_MESSAGE_PROTOCOL_NAME` shouldn't be removed from `EventMeshCommon` because it may be used by SDK clients. ref: https://eventmesh.apache.org/zh/docs/sdk-java/http

Pil0tXia (changes requested):
`EventMeshCommon` class is designed to store constants of EventMesh SDK. It should be kept in `org.apache.eventmesh.client` package although it is referenced by other protocols.

PR #4784, opened 2024-03-05
[ISSUE #3828] Code optimization for StandaloneAdminTest and TestUtils

### LINKED ISSUE #3828: [Enhancement] Do some code optimization.[StandaloneAdminTest TestUtils]
### Search before asking

### Enhancement Request

image
image
image
image
located at:
eventmesh-storage-plugin/eventmesh-storage-standalone/src/test/java/org/apache/eventmesh/storage/standalone/admin/StandaloneAdminTest.java line 73,90,99
eventmesh-storage-plugin/eventmesh-storage-standalone/src/test/java/org/apache/eventmesh/storage/standalone/TestUtils.java line 66
analysis and explanation:
a. This method makes two consecutive calls to the same method, using the same constant parameters, on the same instance, without any intervening changes to the objects. If this method does not make changes to the object, which it appears it doesn't, then making two calls is just a waste. These method calls could be combined by assigning the result into a temporary variable, and using the variable the second time.
b. Method needlessly boxes a boolean constant.
c. Method builds a list from one element using Arrays.asList This method builds a list using Arrays.asList(foo), passing in a single element. Arrays.asList needs to first create an array from this one element, and then build a List that wraps this array. It is simpler to use Collections.singletonList(foo), which does not create the array, and produces a far simpler instance of List. Since both of these arrays are immutable (from the List's point of view) they are equivalent from a usage s tandpoint.

a. Use intermediate variables as storage.
image

b. Use 'Boolean.xxxx'.
image

c. Replace with Collections.singletonList().
image

dr-divago:
Fixes #3828

### Motivation

Optimize code in the following classes : StandaloneAdminTest and TestUtils

### Modifications

Remove duplication
Use native Boolean.TRUE and Boolean.FALSE instead of variable with value true and false
Use Collections.singletonList for creating List

### Documentation

- Does this pull request introduce a new feature?  no
- If yes, how is the feature documented? not applicable
- If a feature is not applicable for documentation, explain why? Optimization in tests

### DIFF
--- a/eventmesh-storage-plugin/eventmesh-storage-standalone/src/test/java/org/apache/eventmesh/storage/standalone/admin/StandaloneAdminTest.java
+++ b/eventmesh-storage-plugin/eventmesh-storage-standalone/src/test/java/org/apache/eventmesh/storage/standalone/admin/StandaloneAdminTest.java
@@ -20,12 +20,11 @@ package org.apache.eventmesh.storage.standalone.admin;
 import static org.apache.eventmesh.storage.standalone.TestUtils.LENGTH;
 import static org.apache.eventmesh.storage.standalone.TestUtils.OFF_SET;
 import static org.apache.eventmesh.storage.standalone.TestUtils.TEST_TOPIC;
-import static org.apache.eventmesh.storage.standalone.TestUtils.TOPIC_DO_NOT_EXISTS;
-import static org.apache.eventmesh.storage.standalone.TestUtils.TOPIC_EXISTS;
 import static org.apache.eventmesh.storage.standalone.TestUtils.createDefaultCloudEvent;
 import static org.apache.eventmesh.storage.standalone.TestUtils.createDefaultMessageContainer;
 import static org.apache.eventmesh.storage.standalone.TestUtils.createDefaultMessageEntity;
 
+import org.apache.eventmesh.api.admin.TopicProperties;
 import org.apache.eventmesh.storage.standalone.broker.StandaloneBroker;
 import org.apache.eventmesh.storage.standalone.broker.model.MessageEntity;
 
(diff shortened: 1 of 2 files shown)


eventmesh-storage-plugin/eventmesh-storage-standalone/src/test/java/org/apache/eventmesh/storage/standalone/admin/StandaloneAdminTest.java
```diff
@@ -90,7 +90,7 @@ public void testDeleteTopic() {
 
     @Test
     public void testGetEvent() throws Exception {
-        Mockito.when(standaloneBroker.checkTopicExist(TEST_TOPIC)).thenReturn(TOPIC_EXISTS);
+        Mockito.when(standaloneBroker.checkTopicExist(TEST_TOPIC)).thenReturn(Boolean.TRUE);
         Mockito.when(standaloneBroker.getMessage(TEST_TOPIC, OFF_SET)).thenReturn(createDefaultCloudEvent());
```
Pil0tXia, line 94: How about adding a `Topic Exists` comment here to compensate the removed boolean member? And a `Topic does not Exist` for L102.

dr-divago: Not sure I understood, checkTopicExist function isn't already conveying exist or not exist in the name?  But I can add a comment no problem

Pil0tXia: Okay, then let's not adding it.

Pil0tXia (approved):

PR #4807, opened 2024-04-02
[ISSUE #4804] Fix SubStreamHandler exception loop by closeOnError

### LINKED ISSUE #4804: [Bug] grpc SubStreamHandler exception
### Search before asking

### Environment

Linux

### EventMesh version

1.10.0

### What happened

There is a problem with the class of SubStreamHandler. grpc is two-way communication. I have integrated grpc pub and sub functions in the project.
This code throws an exception, and every 30 seconds,

 private void senderOnNext(final CloudEvent subscription) {
        try {
            synchronized (sender) {
                sender.onNext(subscription);
            }
        } catch (Exception e) {
            log.error("StreamObserver Error onNext", e);
        }
    }

I analyzed the reason, because when the exception is thrown, you should call onCompleted to close the sender, but that is not done here, resulting in
grpc thinks the connection is still there and keeps throwing exceptions in a loop.

I  tried the method below，it solved the problem。 
Will you change code quickly？   I am using the project，but this problem  stop me。

@Slf4j
public class SubStreamHandler<T> extends Thread implements Serializable {

...
    public SubStreamHandler(final ConsumerServiceStub consumerAsyncClient, final EventMeshGrpcClientConfig clientConfig,
                            final ReceiveMsgHook<T> listener) {
        this.consumerAsyncClient = consumerAsyncClient;
        this.clientConfig = clientConfig;
        this.listener = listener;
    }

    public void sendSubscription(final CloudEvent subscription) {
        synchronized (this) {
            if (this.sender == null) {
                this.sender = consumerAsyncClient.subscribeStream(createReceiver());
            }
        }
        senderOnNext(subscription);
    }

    private StreamObserver<CloudEvent> createReceiver() {
        return new StreamObserver<CloudEvent>() {

            @Override
            public void onNext(final CloudEvent message) {
                T msg = EventMeshCloudEventBuilder.buildMessageFromEventMeshCloudEvent(message, listener.getProtocolType());
                if (msg instanceof Set) {
                    log.info("Received message from Server:{}", message);
                } else {
                    log.info("Received message from Server.|seq={}|uniqueId={}|",
                            EventMeshCloudEventUtils.getSeqNum(message), EventMeshCloudEventUtils.getUniqueId(message));
                    CloudEvent streamReply = null;
                    try {
                        Optional<T> reply = listener.handle(msg);
                        if (reply.isPresent()) {
                            streamReply = buildReplyMessage(message, reply.get());
                        }
                    } catch (Exception e) {
                        log.error("Error in handling reply message.|seq={}|uniqueId={}|",
                                EventMeshCloudEventUtils.getSeqNum(message), EventMeshCloudEventUtils.getUniqueId(message), e);
                    }
                    if (streamReply != null) {
                        log.info("Sending reply message to Server.|seq={}|uniqueId={}|",
                                EventMeshCloudEventUtils.getSeqNum(streamReply), EventMeshCloudEventUtils.getUniqueId(streamReply));
                        senderOnNext(streamReply);
                    }
                }
            }

            @Override
            public void onError(final Throwable t) {
                log.error("Received Server side error", t);
                **close();**
            }

            @Override
            public void onCompleted() {
                log.info("Finished receiving messages from server.");
                close();
            }
        };
    }

    private CloudEvent buildReplyMessage(final CloudEvent reqMessage, final T replyMessage) {
        final CloudEvent cloudEvent = EventMeshCloudEventBuilder.buildEventMeshCloudEvent(replyMessage,
                clientConfig, listener.getProtocolType());

        return CloudEvent.newBuilder(cloudEvent).putAllAttributes(reqMessage.getAttributesMap()).putAllAttributes(cloudEvent.getAttributesMap())
                .putAttributes(ProtocolKey.DATA_CONTENT_TYPE,
                        CloudEventAttributeValue.newBuilder().setCeString(EventMeshDataContentType.JSON.getCode()).build())
                // Indicate that it is a subscription response
                .putAttributes(ProtocolKey.SUB_MESSAGE_TYPE, CloudEventAttributeValue.newBuilder().setCeString(ProtocolKey.SUB_REPLY_MESSAGE).build())
                .build();
    }

    @Override
    public void run() {
        try {
            latch.await();
        } catch (InterruptedException e) {
            log.error("SubStreamHandler Thread interrupted", e);
            Thread.currentThread().interrupt();
        }
    }

    public void close() {
        if (this.sender != null) {
            senderOnComplete();
        }

        latch.countDown();

        log.info("SubStreamHandler closed.");
    }

    private void senderOnNext(final CloudEvent subscription) {
        try {
            synchronized (sender) {
                sender.onNext(subscription);
            }
        } catch (Exception e) {
            log.error("StreamObserver Error onNext", e);
            **close();**
        }
    }

    private void senderOnComplete() {
        try {
            synchronized (sender) {
                sender.onCompleted();
                **sender=null;**
            }
        } catch (Exception e) {
            log.error("StreamObserver Error onComplete", e);
        }
    }
}

### How to reproduce

when exception happen ，this method will cause endless loop。
for example，grpc time out。 
 private void senderOnNext(final CloudEvent subscription) {
        try {
            synchronized (sender) {
                sender.onNext(subscription);
            }
        } catch (Exception e) {
            log.error("StreamObserver Error onNext", e);
        }
    }

### Debug logs

```Java
no
```

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct) *

Pil0tXia:
Fixes #4804

### Motivation

Please see #4804.

### Modifications

- Set the `sender` to `null` in the `senderOnComplete()` method, which requires the `sender` to be recreated in `sendSubscription()` after an error occurs. The only caller of the `senderOnComplete()` method is the `close()` method.
- Add `close()` statement in `onError()` methods.

Existing problem: After calling the `close()` method, the `SubStreamHandler` instance will never run again unless the `EventMeshGrpcConsumer` re-subscribes, which is a known issue.

### Documentation

### DIFF
--- a/eventmesh-sdks/eventmesh-sdk-java/src/main/java/org/apache/eventmesh/client/grpc/consumer/SubStreamHandler.java
+++ b/eventmesh-sdks/eventmesh-sdk-java/src/main/java/org/apache/eventmesh/client/grpc/consumer/SubStreamHandler.java
@@ -40,8 +40,6 @@ public class SubStreamHandler<T> extends Thread implements Serializable {
 
     private final transient CountDownLatch latch = new CountDownLatch(1);
 
-    private volatile boolean isClosed = false;
-
     private final transient ConsumerServiceStub consumerAsyncClient;
 
     private final transient EventMeshGrpcClientConfig clientConfig;
@@ -104,7 +102,6 @@ public class SubStreamHandler<T> extends Thread implements Serializable {
             @Override
             public void onCompleted() {
                 log.info("Finished receiving messages from server.");
-                close();
             }
         };
     }
@@ -132,16 +129,11 @@ public class SubStreamHandler<T> extends Thread implements Serializable {
     }
 
     public void close() {
-        synchronized (sender) {
-            if (isClosed) {
-                return;
-            }
-            if (this.sender != null) {
-                senderOnComplete();
-            }
-            latch.countDown();
-            isClosed = true;
+        if (this.sender != null) {
+            senderOnComplete();
         }
+
+        latch.countDown();
         log.info("SubStreamHandler closed.");
     }
 


mxsm (commented):
@Pil0tXia  Code format used EventMesh code style xml?

Pil0tXia:
@mxsm 

The change in https://github.com/apache/eventmesh/pull/4807/commits/f5898d80531723b2a2fed07b2087ea5716fe1ea9 was made manually.

mxsm (approved):
LGMT

eventmesh-sdks/eventmesh-sdk-java/src/main/java/org/apache/eventmesh/client/grpc/consumer/SubStreamHandler.java
```diff
@@ -145,13 +146,15 @@ private void senderOnNext(final CloudEvent subscription) {
...
             log.error("StreamObserver Error onNext", e);
+            close();
         }
     }
 
     private void senderOnComplete() {
         try {
             synchronized (sender) {
                 sender.onCompleted();
+                sender = null;
```
pandaapo, line 157: Isn't this question a bit ...... You set the lock object to `null` in the synchronization block, wouldn't it be too easy for other threads waiting for the lock to access `null`?

这里的问题是不是有点……您在同步块中将锁对象置为`null`，那是不是太容易出现其他等待锁的线程访问到`null`了？

Pil0tXia: If the lock object is set to null within the synchronized block, it does not affect the threads that are already waiting for this lock. This is because when a thread enters a synchronized block, it actually obtains a lock on the monitor of the lock object, not on the object reference itself.

Therefore, even if `sender` is set to null within the synchronized block, other threads that are waiting for the `sender` lock will continue to wait until the current thread exits the synchronized block and releases the monitor lock on `sender`.

However, setting the lock object to null may cause subsequent synchronization operations to fail, because they may attempt to acquire a lock on a null reference, which will throw a `NullPointerException`.

In this case, the only caller of the `senderOnComplete()` method is the `close()` method, which may not be that easy to encounter synchronization failure. However, I'm still wondering that if there is a better way to resolve this issue.

pandaapo: > If the lock object is set to null within the synchronized block, it does not affect the threads that are already waiting for this lock. This is because when a thread enters a synchronized block, it actually obtains a lock on the monitor of the lock object, not on the object reference itself.
> 
> Therefore, even if `sender` is set to null within the synchronized block, other threads that are waiting for the `sender` lock will continue to wait until the current thread exits the synchronized block and releases the monitor lock on `sender`.
> 
> However, setting the lock object to null may cause subsequent synchronization operations to fail, because they may attempt to acquire a lock on a null reference, which will throw a `NullPointerException`.

This is the bad problem that this code will cause.

> In this case, the only caller of the `senderOnComplete()` method is the `close()` method, which may not be that easy to encounter synchronization failure. 

This is not a reason for this coding. Writing concurrent code and expecting no concurrency because it will definitely cause error. If so, concurrency programming would have no robustness at all. 

> wondering that if there is a better way to resolve this issue.

Of course it can be resolved correctly, and must be. 
For example, One approach is to add a flag to control that `close()` only calls `senderOnComplete()` once. Another approach is to optimize this circular logic `close() > senderOnComplete() > onCompleted() > close() > ...`. Analyzing the code reveals that `onCompleted()` does not need to call `close()` again because the key of `close()` method is `latch.countDown()`, so there is no need to go around in a large circle to execute `latch.countDown()` again.

Pil0tXia: There're multiple callers of `close()`, so adding a flag would be more accurate and easier to manage. Fixed in https://github.com/apache/eventmesh/pull/4807/commits/774397fcae4905dc0ec8230eab18bf655f202e8f.

eventmesh-sdks/eventmesh-sdk-java/src/main/java/org/apache/eventmesh/client/grpc/consumer/SubStreamHandler.java
```diff
@@ -129,6 +132,12 @@ public void run() {
     }
 
     public void close() {
+        // Avoid closing multiple times
+        if (isClosed) {
+            return;
+        }
+        isClosed = true;
```
pandaapo, line 139: The way you add the flag doesn't solve the problem of multiple closing under concurrency.

eventmesh-sdks/eventmesh-sdk-java/src/main/java/org/apache/eventmesh/client/grpc/consumer/SubStreamHandler.java
pandaapo: Regardless of the approach ultimately chosen, please remember to remove this line of bad code `sender = null`.

Pil0tXia: Although this statement is requested by issue author, I think his issue can still be resolved after removing this statement.

Fixed in https://github.com/apache/eventmesh/pull/4807/commits/d0ac6c0900a7be661f8a3e434ae364c42ba56c96

eventmesh-sdks/eventmesh-sdk-java/src/main/java/org/apache/eventmesh/client/grpc/consumer/SubStreamHandler.java
Pil0tXia: The `onCompleted` method of `StreamObserver` is idempotent. After the first call to the `onCompleted` method, the stream is considered to be completed, and any subsequent `onNext` or `onCompleted` calls will be ignored.

Since the `sender == null` statement has been removed, so I reverted the close flag in https://github.com/apache/eventmesh/pull/4807/commits/878009a99b1b64ffdec0b579bb564b8c9e31afec.

eventmesh-sdks/eventmesh-sdk-java/src/main/java/org/apache/eventmesh/client/grpc/consumer/SubStreamHandler.java
pandaapo: Although this statement is requested by issue author, but we should be responsible for the correctness before copying the issue author's code.

eventmesh-sdks/eventmesh-sdk-java/src/main/java/org/apache/eventmesh/client/grpc/consumer/SubStreamHandler.java
pandaapo: > The `onCompleted` method of `StreamObserver` is idempotent. After the first call to the `onCompleted` method, the stream is considered to be completed, and any subsequent `onNext` or `onCompleted` calls will be ignored.

Are you sure that grpc just ignores subsequent calls rather than throwing an exception that could be and should be avoided?
> Since the `sender == null` statement has been removed, so I reverted the close flag in [878009a](https://github.com/apache/eventmesh/commit/878009a99b1b64ffdec0b579bb564b8c9e31afec).

The removal of `sender == null` is not the main reason for reverting this newly added close flag. The main reason is that this close flag was used in a wrong way for concurrent programming.

Pil0tXia: Sorry, `onCompleted` doesn't throw an exception, `onNext` does. I forgot to remove `onNext` after pasting the text.

There's only `onCompleted` call here, which should not throw an exception if `onCompleted` itself didn't throw one.

Is there anything I should change?

pandaapo: > `onCompleted` doesn't throw an exception (when called again)

Could you please provide the basis for this conclusion? Was it based on testing with multiple calls, or some other evidence? Because from the grpc source code, it seems that it cannot explain why a secondary calling `onCompleted` would not throw an exception.

Pil0tXia: https://github.com/grpc/grpc-java/blob/497e1552172632107574d67d7c092a3f5a39a980/stub/src/main/java/io/grpc/stub/ServerCalls.java#L393-L397

When `onCompleted` is called, it will set `completed` flag to true.

https://github.com/grpc/grpc-java/blob/497e1552172632107574d67d7c092a3f5a39a980/stub/src/main/java/io/grpc/stub/ServerCalls.java#L359-L381

When `onNext` is called, it will check `completed` flag and throw an exception.

Anyway, I added a check in https://github.com/apache/eventmesh/pull/4807/commits/8e812cdcb68a7a1e90241b30739cba77998b82e5 to keep `senderOnComplete` called once.

pandaapo: > https://github.com/grpc/grpc-java/blob/497e1552172632107574d67d7c092a3f5a39a980/stub/src/main/java/io/grpc/stub/ServerCalls.java#L393-L397
> 
> When `onCompleted` is called, it will set `completed` flag to true, but won't throw exceptions.

I suggest you take another look inside `call.close(Status.OK, new Metadata())` line of this piece of source code.

eventmesh-sdks/eventmesh-sdk-java/src/main/java/org/apache/eventmesh/client/grpc/consumer/SubStreamHandler.java
```diff
@@ -129,12 +130,15 @@ public void run() {
     }
 
     public void close() {
-        if (this.sender != null) {
-            senderOnComplete();
+        synchronized (sender) {
+            if (latch.getCount() == 0) {
+                return;
+            }
```
pandaapo, line 136: These concurrent programming codes add another layer of `synchronized` outside of `senderOnComplete()` and `CountDownLatch`. This is equivalent to adding another lock outside of the `synchronized` lock, and it wastes the performance advantages of `CountDownLatch` over locks. 
This is because you added `CountDownLatch#getCount()`, which destroys the atomicity of `CountDownLatch` in the original code. The deeper reason for this is the unnecessary circular logic `close() > senderOnComplete() > onCompleted() > close() > ...` that I mentioned earlier.

eventmesh-sdks/eventmesh-sdk-java/src/main/java/org/apache/eventmesh/client/grpc/consumer/SubStreamHandler.java
Pil0tXia: Thanks for your reminding, it will throw an exception.

eventmesh-sdks/eventmesh-sdk-java/src/main/java/org/apache/eventmesh/client/grpc/consumer/SubStreamHandler.java
Pil0tXia: >another lock outside of the `synchronized` lock

`senderOnComplete()` may be called by another caller in the future.

>`CountDownLatch#getCount()` destroys the atomicity of `CountDownLatch` in the original code

The only usage of `latch.countDown()` is included in synchronized block with `latch.getCount() == 0`, so it can be considered atomic. Well maybe we can't prevent somebody call `countDown()` again somewhere else either, so it may be a potential problem.

>optimize this circular logic `close() > senderOnComplete() > onCompleted() > close() > ...`. Analyzing the code reveals that `onCompleted()` does not need to call `close()` again because the key of `close()` method is `latch.countDown()`, so there is no need to go around in a large circle to execute `latch.countDown()`.

You mean replace the `close()` call of `onCompleted()` with a `latch.countDown()`?

I used boolean to prevent latch called by somebody else in https://github.com/apache/eventmesh/pull/4807/commits/4d8694898e20210aced416f71349c4a220257331. Does that sounds good to you?

pandaapo: > > another lock outside of the `synchronized` lock (in `senderOnComplete()`)
> 
> `senderOnComplete()` may be called by another caller in the future.

The `synchronized` inside `senderOnComplete()` can ensures safety.

> > `CountDownLatch#getCount()` destroys the atomicity of `CountDownLatch` in the original code
>
> The only usage of `latch.countDown()` is included in synchronized block with `latch.getCount() == 0`, so it can be considered atomic. 

Regarding "`latch.getCount()==0` + `latch.countDown()` can be considered atomic", you can think about it again.

> Well maybe we can't prevent somebody call `countDown()` again somewhere else either, so it may be a potential problem.

When `CountDownLatch` counter reaches 0 and `countDown()` is called again, nothing happens, no exceptions are thrown, and there are no potential problems.

>>optimize this circular logic `close() > senderOnComplete() > onCompleted() > close() > ...`. Analyzing the code reveals that `onCompleted()` does not need to call `close()` again because the key of `close()` method is `latch.countDown()`, so there is no need to go around in a large circle to execute `latch.countDown()`.
>
> You mean replace the `close()` call of `onCompleted()` with a latch.countDown()?

The meaning of this previously mentioned approach is to remove `close()` from `onCompleted()`.

eventmesh-sdks/eventmesh-sdk-java/src/main/java/org/apache/eventmesh/client/grpc/consumer/SubStreamHandler.java
```diff
@@ -129,12 +132,16 @@ public void run() {
     }
 
     public void close() {
-        if (this.sender != null) {
-            senderOnComplete();
+        synchronized (sender) {
+            if (isClosed) {
+                return;
```
pandaapo, line 137: Firstly, setting aside the context, when this flag variable is wrapped by `synchronized`, its `volatile` is redundant and even adds overhead.

Returning to the context, as explained earlier, there's no need to add another `synchronized` outside of `senderOnComplete()`'s lock. If use the approach previously mentioned of adding a flag, consider using `AtomicBoolean`.

eventmesh-sdks/eventmesh-sdk-java/src/main/java/org/apache/eventmesh/client/grpc/consumer/SubStreamHandler.java
Pil0tXia: May I ask that why a synchronized block with `latch.getCount() == 0` and `latch.countDown()` is not considered atomic?

>this previously mentioned approach is to remove `close()` from `onCompleted()`.

Done in https://github.com/apache/eventmesh/pull/4807/commits/dbabca713d2fe7f1d7b58bf4bc2e371a00a1cc0a.

eventmesh-sdks/eventmesh-sdk-java/src/main/java/org/apache/eventmesh/client/grpc/consumer/SubStreamHandler.java
Pil0tXia: I think a synchronized block is necessary to keep `isClosed` flag judgment and operation atomic. Even with `AtomicBoolean`, atomicity is only guaranteed for a single operation. Without the synchronized code block, it is possible for the second thread to pass the `isClosed == false` judgment before `isClosed` is set to true.

eventmesh-sdks/eventmesh-sdk-java/src/main/java/org/apache/eventmesh/client/grpc/consumer/SubStreamHandler.java
pandaapo: > May I ask that why a synchronized block with `latch.getCount() == 0` and `latch.countDown()` included is not considered atomic?

A synchronized block is naturally atomic. The original sentence is "added `CountDownLatch#getCount()` destroys the atomicity of `CountDownLatch` in the original code", so "`latch.getCount() == 0` + `latch.countDown()`" has to be wrapped with another synchronized.

eventmesh-sdks/eventmesh-sdk-java/src/main/java/org/apache/eventmesh/client/grpc/consumer/SubStreamHandler.java
pandaapo: > I think a synchronized block is necessary to keep `isClosed` flag judgment and operation atomic. Even with `AtomicBoolean`, atomicity is only guaranteed for a single operation. Without the synchronized code block, it is possible for the second thread to pass the `isClosed == false` judgment before `isClosed` is set to true.

You can learn about some methods of `AtomicBoolean`.

pandaapo (approved):

mxsm (approved):

PR #5102, opened 2024-09-08
[ISSUE #5101] Define and standardize some common configurations for all Sources

### LINKED ISSUE #5101: [Enhancement] Define and standardize some common configurations for all Sources
### Search before asking

### Enhancement Request

In the various implementations of `Source`, there are often some common configurations scattered across different places, such as the capacity of the storage queue, the number of elements to poll each time, etc. Additionally, in some source implementations, the `poll` method can block for up to several seconds in the worst-case scenario, reaching `maxBatchSize * pollTimeout`. Therefore, it is necessary to define common parameters that support configuration and also limit the maximum waiting time for `poll`.

1. Added a `PollConfig` configuration class that includes parameters for `capacity`, `maxBatchSize`, and `maxWaitTime`, and incorporated it as a variable in `SourceConfig` to support user configuration with default values.
2. Replaced the fixed constants for `capacity`, `maxBatchSize`, and `maxWaitTime` in the original source implementations with the corresponding variables from `PollConfig`.
3. Optimized some `poll` methods to ensure the maximum waiting time does not exceed `maxWaitTime`, preventing blocking for several seconds.

### Code of Conduct

cnzakii:
Fixes #5101 

### Motivation

For further details, refer to  #5101 

### Modifications

For further details, refer to  #5101 

### Documentation

### DIFF
--- a/eventmesh-common/src/main/java/org/apache/eventmesh/common/config/connector/mq/kafka/SourceConnectorConfig.java
+++ b/eventmesh-common/src/main/java/org/apache/eventmesh/common/config/connector/mq/kafka/SourceConnectorConfig.java
@@ -32,5 +32,4 @@ public class SourceConnectorConfig {
     private String enableAutoCommit = "false";
     private String sessionTimeoutMS = "10000";
     private String maxPollRecords = "1000";
-    private int pollTimeOut = 100;
 }
--- a/eventmesh-common/src/main/java/org/apache/eventmesh/common/config/connector/PollConfig.java
+++ b/eventmesh-common/src/main/java/org/apache/eventmesh/common/config/connector/PollConfig.java
@@ -17,34 +17,27 @@
 
 package org.apache.eventmesh.common.config.connector;
 
-public class Constants {
+import lombok.Data;
 
-    public static final String ENV_TARGET = "connectorTarget";
-
-    public static final String ENV_PORT = "connectorPort";
-
-    public static final String ENV_SOURCE_CONFIG_FILE = "sourceConnectorConf";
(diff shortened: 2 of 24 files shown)


xwm1992 (approved):

eventmesh-common/src/main/java/org/apache/eventmesh/common/config/connector/mq/kafka/SourceConnectorConfig.java
```diff
@@ -32,5 +32,4 @@ public class SourceConnectorConfig {
     private String enableAutoCommit = "false";
     private String sessionTimeoutMS = "10000";
     private String maxPollRecords = "1000";
-    private int pollTimeOut = 100;
 }
```
Pil0tXia, line 35: Can we allocate an appropriate timeout value for a source where we anticipate the RT might be significantly slower, and override the global default value?

If this feature isn't supported yet, you can implement it in a future PR. This is very important for users who use multiple sources with the same configuration files.

eventmesh-sdks/eventmesh-sdk-c/third_party/curl
Pil0tXia, line 1: Did the significant changes in these two git submodules come from upstream updates? Are these changes compatible with our C SDK? If untested, it's advisable not to update the versions of third-party dependencies, because the C SDK cannot currently be tested in CI (since it cannot be compiled at the moment).

Pil0tXia (approved):

xwm1992 (approved):

Issue #5296, opened 2026-08-18
[Architecture Review] Consolidate EventMesh boundaries and production-grade consistency


qqeasonchen:
# Architecture Review: Consolidate EventMesh boundaries and production-grade consistency

## Summary

EventMesh's core direction is sound, but the biggest risk today is that, outside the "unified Runtime", multiple parallel stacks still coexist for protocols, state, clustering, and task handling. The priority is to **converge boundaries and reliability semantics before adding more features**.

This issue is the parent of a series of focused work items. See the **Execution plan** below for the machine-readable ordering and dependency graph.

### P0 (must address first)
- [ ] #5297 — Decouple `eventmesh-runtime` responsibilities via internal packages
- [ ] #5298 — Partition the `eventmesh-common` module into internal packages (model / wire / config / util)
- [ ] #5299 — Single protocol path (HTTP + CloudEvents) and ingress/egress adapters
- [ ] #5301 — Unified state control-plane stores
- [ ] #5293 — (pre-existing) cluster delivery topology + fence stale partition owners

### P1 (plan soon)
- [ ] #5302 — A2A must not form a parallel Runtime
- [ ] #5303 — Standardize the Storage SPI capability model + TCK
- [ ] #5304 — Unified multi-tenant / security / quota entrypoint

### P2 (engineering governance)
- [ ] #5305 — Gradle build + module dependency guardrails
- [ ] #5306 — Sync documentation narrative with implementation status

> Note: #5300 (DeliveryTopology) was folded into #5293 and closed as a duplicate.

## Execution plan (machine-readable)

> **Vocabulary for agents**
> - `Severity` = problem importance (`P0` / `P1` / `P2`).
> - `Phase` = execution order (`0` .. `4`); **orthogonal to severity** — Phase 0 runs first even if its severity is lower.
> - `Depends on` = must land before this issue; `Blocks` = cannot start until this issue lands.
> - Issues are referenced by number (`#NNNN`) so they can be resolved programmatically.

### Phase 0 — Guardrails & direction (no dependencies; start immediately)
- [ ] #5305 — dependency guardrails (severity P2, scheduled first: cheap, prevents further coupling)
- [ ] #5293 — decide the cluster `DeliveryTopology` (severity P1)

### Phase 1 — Single-instance at-least-once correctness (stop the bleeding; independent, parallelizable)
- [ ] #5289 — atomic monotonic offset writes
- [ ] #5290 — do not retire deliveries on offset-persist failure
[...]

Issue #5295, opened 2026-08-18
[Bug] Gate RocketMQ 5 POP broker ACK on distribution completion


qqeasonchen:
> **Execution meta** — see plan: [#5296](https://github.com/apache/eventmesh/issues/5296)
> - **Phase:** `2 — unified state control plane`
> - **Severity:** `P0`
> - **Depends on:** #5301
> - **Blocks:** (none)

## Problem

RocketMQ 5 POP messages defer broker ACK until `ReliableDispatcher.ack()` invokes the callback associated with `empopck`. A single POP check key can be attached to multiple Runtime deliveries when a message is broadcast or multicast. The first client ACK removes the callback and ACKs the broker, even if other required targets have not ACKed.

Relevant code:

- `eventmesh-storage-plugin/eventmesh-storage-rocketmq5/src/main/java/org/apache/eventmesh/storage/rocketmq5/storage/RocketMQ5RemotingStoragePlugin.java`: `pendingPopAcks` and `ackPulledMessage()`
- `eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/ingress/UniIngressService.java`: creation of one `mqAck` callback from `empopck`
- `eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/delivery/ReliableDispatcher.java`: execution of the callback after client ACK

## Reproduction

1. Use RocketMQ 5 POP and create two active `BROADCAST` subscribers.
2. Publish one message and wait until both receive it.
3. ACK only subscriber A.
4. Verify that the broker ACK has not yet occurred.
5. Crash the Runtime before subscriber B ACKs.
6. The broker must redeliver to satisfy B's unfinished obligation.

## Expected behavior

The broker ACK must occur only when the delivery-completion rule for the distribution mode is met:

- LOAD_BALANCE: selected target ACKs.
- BROADCAST: all required active targets ACK.
- MULTICAST: the configured target/quorum rule is satisfied.

## Proposed direction

Introduce a broker-ACK barrier keyed by POP check key. It should track required delivery IDs, acknowledged delivery IDs, terminal policy, and whether the broker ACK has already been executed. The barrier must be idempotent and compatible with restart/failure recovery.

## Acceptance criteria

- [ ] In BROADCAST mode, the first of multiple client ACKs does not ACK the POP message at the broker.
- [ ] The final required ACK executes exactly one broker ACK.
- [ ] Duplicate and out-of-order ACKs do not execute multiple broker ACKs.
[...]

Issue #5294, opened 2026-08-18
[Bug] Preserve unacknowledged broadcast deliveries during cursor recovery


qqeasonchen:
> **Execution meta** — see plan: [#5296](https://github.com/apache/eventmesh/issues/5296)
> - **Phase:** `2 — unified state control plane`
> - **Severity:** `P0`
> - **Depends on:** #5301
> - **Blocks:** (none)

## Problem

The Runtime records one physical MQ cursor per `topic + partition` under the reserved `__mqcursor__` client ID whenever any delivery is ACKed. In broadcast or multicast delivery, the same physical message is delivered to multiple targets. A fast target can advance this shared cursor while another target still has an unacknowledged copy.

On restart, `UniRuntime.alignPullOffsetsToAck()` uses the shared cursor to rewind Kafka/RocketMQ 4 pull state. The model cannot represent per-subscriber completion obligations, so a slow target's unacknowledged copy can be skipped or replay behavior can become backend-dependent.

Relevant code:

- `eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/ingress/UniIngressService.java`: `pullAndDispatchPartition()`
- `eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/delivery/ReliableDispatcher.java`: `ack(String)`
- `eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/boot/UniRuntime.java`: `alignPullOffsetsToAck()`

## Reproduction

1. Use Kafka or RocketMQ 4 with a topic/partition containing one message.
2. Register two `BROADCAST` subscribers, A and B.
3. Deliver the message to both subscribers.
4. ACK from A only; leave B unacknowledged.
5. Crash/restart the Runtime before B ACKs.
6. Verify whether B receives the message again. The required result must not depend on which subscriber ACKed first.

## Expected behavior

For broadcast/multicast, recovery must preserve every uncompleted delivery obligation. A physical cursor must not advance past a message until the configured completion rule for that distribution mode is satisfied.

## Proposed direction

- Define recovery semantics independently for `LOAD_BALANCE`, `BROADCAST`, and `MULTICAST`.
- Track physical progress per subscription/group, or persist delivery obligations keyed by topic, physical partition, physical offset, subscription/client, and completion state.
- Derive restart rewind from the minimum incomplete physical position rather than a single global `__mqcursor__` value.
- Document the duplicate-delivery bound for each mode.

## Acceptance criteria

[...]

Issue #5288, opened 2026-08-18
[Bug] Start and manage membership heartbeats in sticky cluster mode


qqeasonchen:
> **Execution meta** — see plan: [#5296](https://github.com/apache/eventmesh/issues/5296)
> - **Phase:** `3 — cluster model landing`
> - **Severity:** `P0`
> - **Depends on:** #5293
> - **Blocks:** (none)

## Problem

`EventMeshApplication.enableCluster()` configures the sticky membership model and creates `ClusterMembership`, but does not visibly start a periodic membership heartbeat. The periodic call to `ClusterMembership.heartbeat()` is currently driven by `PartitionOwnership.refresh()`, while the sticky path intentionally does not wire `PartitionOwnership`.

As a result, the default sticky-cluster path may not publish/refresh `/em/instances/<id>` membership records. This affects `/session/recommend`, load-based instance selection, stale-member expiry, and graceful leave semantics.

Relevant code:

- `eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/boot/EventMeshApplication.java`: `enableCluster()`
- `eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/cluster/ClusterMembership.java`: `heartbeat()`
- `eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/cluster/PartitionOwnership.java`: heartbeat scheduling in `refresh()`

## Reproduction

1. Start `EventMeshApplication` with clustered Meta configuration.
2. Call `enableCluster()` but do not configure `PartitionOwnership`.
3. Inspect MetaStore for `/em/instances/<instanceId>` over multiple TTL intervals.
4. Query `/session/recommend` from another instance.
5. Observe whether membership and load information is published and refreshed.

## Expected behavior

Every cluster topology that depends on membership must own an explicit membership lifecycle: immediate startup heartbeat, periodic refresh before TTL expiry, health/degraded behavior when Meta is unavailable, and deletion on graceful shutdown.

## Proposed direction

- Introduce a dedicated membership lifecycle service/scheduler independent of partition ownership.
- Start it from `enableCluster()` for sticky mode.
- Stop it before `leave()` during shutdown.
- Define Meta outage behavior for readiness and `/session/recommend`.

## Acceptance criteria

- [ ] Sticky mode writes a membership record immediately after startup.
- [ ] The record timestamp and load snapshot refresh periodically before TTL expiry.
- [ ] Graceful shutdown removes the membership record.
[...]

PR #5307, opened 2026-08-19
fix(reliability): atomic offset writes, durable DLQ, epoch delivery IDs (#5289 #5290 #5291 #5292)

### LINKED ISSUE #5289: [Bug] Make RocksDB and Meta offset writes atomically monotonic
> **Execution meta** — see plan: [#5296](https://github.com/apache/eventmesh/issues/5296)
> - **Phase:** `1 — single-instance correctness`
> - **Severity:** `P1`
> - **Depends on:** (none)
> - **Blocks:** #5301

## Problem

`RocksDBOffsetStore.writeOffset()` implements monotonic writes as `get → compare → put`. This is not atomic. Concurrent ACKs for the same offset key can interleave so that a lower offset overwrites a higher offset.

`MetaBackedOffsetStore` also stores one dirty value per key without a documented monotonic merge at the remote flush boundary.

Relevant code:

- `eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/offset/RocksDBOffsetStore.java`: `writeOffset()`
- `eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/cluster/MetaBackedOffsetStore.java`: `writeOffset()` and `flushDirty()`

## Reproduction

1. Initialize one `RocksDBOffsetStore` key at offset 10.
2. Use two threads synchronized by a barrier.
3. Concurrently write offsets 11 and 20 for the same topic/client/partition.
4. Force the lower writer to perform its final `put` after the higher writer.
5. Observe that the stored value can become 11.

## Expected behavior

Offsets must be monotonic per key under concurrent writes. Once offset 20 is durably visible, no later concurrent operation may reduce it to 11.

## Proposed direction

Use one of:

- per-key synchronization around read/compare/write;
- a RocksDB merge operator with max semantics;
- a serialized state writer; or
- append-only completion state plus deterministic compaction.

Apply equivalent monotonic behavior to local and remote/meta tiers.

## Acceptance criteria

- [ ] Concurrent writes for one key never decrease the persisted offset.
- [ ] Stress tests repeatedly exercise competing lower/higher writes.
- [ ] Local RocksDB and Meta-backed modes preserve monotonicity.
- [ ] Failures and retries cannot overwrite a newer offset with an older value.
- [ ] The implementation documents its concurrency guarantee.

## Suggested labels

`bug`, `reliability`, `storage`, `priority:p1`

### LINKED ISSUE #5290: [Bug] Do not retire deliveries when durable offset persistence fails
> **Execution meta** — see plan: [#5296](https://github.com/apache/eventmesh/issues/5296)
> - **Phase:** `1 — single-instance correctness`
> - **Severity:** `P0`
> - **Depends on:** (none)
> - **Blocks:** #5301

## Problem

`RocksDBOffsetStore.writeOffset()` catches persistence errors and only increments a failure counter. `ReliableDispatcher.ack()` does not receive a write result, so it removes the pending delivery and can execute a RocketMQ 5 broker ACK even if the subscriber offset or physical cursor was not durably persisted.

Relevant code:

- `eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/offset/RocksDBOffsetStore.java`: `writeOffset()`
- `eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/delivery/ReliableDispatcher.java`: `ack(String)`

## Reproduction

1. Use an `OffsetStore` test double that fails `writeOffset`.
2. Deliver a message with a deferred broker ACK callback.
3. ACK the delivery.
4. Observe that the dispatcher retires the delivery and may invoke broker ACK despite offset persistence failure.
5. Restart from a pull cursor that is ahead of the durable ACK state.

## Expected behavior

An ACK must not be considered successful until required durable state is written. If offset/delivery-state persistence fails, EventMesh must not execute source broker ACK and must preserve a retryable delivery state.

## Proposed direction

- Change the offset persistence contract to return a result or throw a typed exception.
- Make `ReliableDispatcher.ack()` transactional in intent: validate → persist completion/offset → broker ACK → retire delivery.
- On persistence failure, retain or requeue the delivery and return a retryable error.
- Add health/metrics for durable ACK failures.

## Acceptance criteria

- [ ] A failed offset write prevents broker ACK callback execution.
- [ ] A failed offset write does not permanently retire the delivery.
- [ ] The ACK endpoint reports failure/retryability to the client.
- [ ] Restart after an offset-write failure does not skip the message.
- [ ] Metrics expose durable ACK/offset write failures.
- [ ] Tests cover Kafka/RocketMQ 4 cursor recovery and RocketMQ 5 deferred broker ACK behavior.

## Suggested labels

`bug`, `reliability`, `storage`, `priority:p0`

### LINKED ISSUE #5291: [Bug] Prevent stale ACKs from matching reused delivery IDs
> **Execution meta** — see plan: [#5296](https://github.com/apache/eventmesh/issues/5296)
> - **Phase:** `1 — single-instance correctness`
> - **Severity:** `P0`
> - **Depends on:** (none)
> - **Blocks:** #5301

## Problem

`ReliableDispatcher` generates delivery IDs from an in-memory sequence (`d-<n>`). The sequence restarts for every Runtime process and is independent on every instance. The ACK API resolves delivery state using only `deliveryId`.

A delayed ACK from a pre-restart delivery can therefore collide with a newly created delivery using the same ID. The stale ACK can retire the new delivery, advance its offset, and—on RocketMQ 5 POP—trigger its deferred broker ACK.

Relevant code:

- `eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/delivery/ReliableDispatcher.java`: `nextDeliveryId()` and `ack(String)`
- `eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/push/PushService.java`: callback lookup by `deliveryId`

## Reproduction

1. Start a Runtime and subscribe a manual-ACK client.
2. Deliver message A and retain its `deliveryId` without ACKing it.
3. Restart the Runtime or route the client to another Runtime instance.
4. Deliver message B so the new Runtime creates the same sequence-based `deliveryId`.
5. Submit the delayed ACK for message A.
6. Observe that message B can be acknowledged even though the client never processed it.

## Expected behavior

An ACK must only complete the exact delivery issued by the same Runtime delivery epoch and subscription/client context. Stale ACKs must be rejected without advancing offsets or executing a broker ACK callback.

## Proposed direction

- Replace process-local sequence-only IDs with an opaque globally unique delivery token containing a boot/instance epoch plus a random or globally unique component.
- Bind ACK validation to delivery ownership, at minimum `clientId` and subscription context; add an explicit delivery epoch/token if the HTTP API can carry it.
- Preserve idempotency for duplicate ACKs of the same current delivery.
- Return a distinguishable stale/unknown ACK result for observability.

## Acceptance criteria

- [ ] A Runtime restart cannot reuse a delivery token issued before the restart.
- [ ] Two Runtime instances cannot generate colliding delivery tokens.
- [ ] A stale ACK after restart/instance switch does not remove a new pending delivery.
- [ ] A stale ACK does not advance a subscriber offset, MQ cursor, or RocketMQ 5 deferred broker ACK.
- [ ] Duplicate ACK for the same completed delivery remains safe and idempotent.
- [ ] Unit tests cover restart, multi-instance, delayed ACK, and duplicate ACK cases.

## Suggested labels

`bug`, `reliability`, `priority:p0`

### LINKED ISSUE #5292: [Bug] Make DLQ transition durable before retiring source delivery
> **Execution meta** — see plan: [#5296](https://github.com/apache/eventmesh/issues/5296)
> - **Phase:** `1 — single-instance correctness`
> - **Severity:** `P0`
> - **Depends on:** (none)
> - **Blocks:** #5301

## Problem

When a delivery exhausts retries, `ReliableDispatcher.tick()` removes it from the in-memory pending map and invokes `DeadLetterSink`. The Runtime DLQ sink publishes asynchronously to `<topic>_DLQ`; a send failure is logged but does not restore the original delivery, persist an outbox record, or retry the terminal transition.

For RocketMQ 5 POP, the current DLQ path intentionally does not invoke the deferred source broker ACK. After POP invisible time expires, the source message can be redelivered and dead-lettered again. If the source ACK is added without a durable DLQ write boundary, a DLQ write failure can lose the source message.

Relevant code:

- `eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/delivery/ReliableDispatcher.java`: `tick()` terminal DLQ path
- `eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/ingress/UniIngressService.java`: `deadLetterSink()`
- `eventmesh-storage-plugin/eventmesh-storage-rocketmq5/src/main/java/org/apache/eventmesh/storage/rocketmq5/storage/RocketMQ5RemotingStoragePlugin.java`: deferred POP ACK handling

## Reproduction

1. Configure a low retry budget and a storage double that fails writes to `<topic>_DLQ`.
2. Deliver a message and allow retry exhaustion.
3. Observe that the original pending delivery is already gone despite failed DLQ persistence.
4. For RocketMQ 5 POP, allow a successful DLQ write but no source ACK, then wait beyond invisible time.
5. Observe source redelivery and duplicate DLQ records.

## Expected behavior

DLQ transition must be durable and idempotent:

- A source message is terminal only after DLQ persistence succeeds.
- A failed DLQ write leaves the source retryable/recoverable.
- A POP source is broker-ACKed only after durable DLQ completion.
- Repeated source redelivery does not create duplicate logical DLQ records.

## Proposed direction

- Make `DeadLetterSink` return a completion result rather than `void`.
- Introduce a durable terminal-transition/outbox record keyed by source message identity.
- Persist DLQ metadata: original topic, source message ID, physical location or POP key, attempts, failure reason, timestamps, and replay count.
- After durable DLQ completion, execute the source broker ACK exactly once where required.

## Acceptance criteria

- [ ] DLQ send failure cannot silently retire the original delivery.
- [ ] The Runtime retries or durably records a failed DLQ transition.
- [ ] RocketMQ 5 POP source messages are ACKed only after successful durable DLQ write.
- [ ] A POP message that reaches DLQ does not reappear and create duplicate DLQ records after invisible time.
- [ ] DLQ replay preserves and exposes source metadata.
- [ ] Unit tests and RocketMQ 5 E2E tests cover DLQ write failure, source redelivery, idempotency, and replay.

## Suggested labels

`bug`, `reliability`, `storage`, `priority:p0`

qqeasonchen:
## Motivation

This PR closes the remaining Phase 1 reliability gaps identified in the architecture audit (#5296). The `lite-topic` / `EventMeshFrame` architecture landed in 0be4e83 already covers #5288, #5294 (mq cursor recording) and #5295 (deferred broker ACK via `ackPulledMessage` + `mqAckCallback`); this change fixes the four issues that remained open.

### #5289 — Offset writes were not atomic (regression window)
`RocksDBOffsetStore.writeOffset` now performs its read-compare-put under a write lock, so a concurrently arriving ACK carrying a lower offset can no longer regress the stored offset between the read and the put.

### #5290 — Deliveries retired even when offset persistence failed
`OffsetStore.writeOffset` now returns `boolean`. `ReliableDispatcher.ack()` only retires a delivery after the durable offset write succeeds; on failure the delivery stays in flight with backoff and redelivery, instead of being silently dropped. `InMemoryOffsetStore` keeps monotonicity via `AtomicLong.accumulateAndGet(offset, Math::max)`; `MetaBackedOffsetStore` propagates the local store's result.

### #5291 — Delivery IDs reused across restarts / instances
Delivery IDs now embed the process boot epoch and a per-process random salt (`bootEpoch + instanceSalt + seq`), so a stale ACK arriving after a restart can never alias a fresh delivery.

### #5292 — DLQ writes were fire-and-forget (silent loss)
`DeadLetterSink.deadLetter` now returns `CompletableFuture<Boolean>`. `ReliableDispatcher.tick()` retires a dead-lettered delivery only after storage confirms the DLQ write; otherwise the delivery returns to pending and the dead-letter transition is retried. `UniIngressService` completes the future from the storage send callback.

## Documentation

- Does this PR introduce a new feature? No, reliability bug fixes only.

## Test Plan

- Updated `ReliableDispatcherTest`, `DeferredAckDispatcherTest`, `DeliveryVolatileAndResurrectTest`, `MqCursorRecordingTest` for the new `DeadLetterSink` signature.
- Full `eventmesh-runtime` test suite: **162 tests, all green**.
- CI green on ubuntu-latest + macOS-latest (JDK 21): https://github.com/qqeasonchen/eventmesh/actions/runs/32206684906

Closes #5289
Closes #5290
Closes #5291
Closes #5292
Refs #5296


qqeasonchen:
Note on the failed `License Check` workflow: the 23 files it reports (e.g. `eventmesh-agent/conf/agent.properties`, `eventmesh-connector-plugin/*/build.gradle`, `resources/logo.svg`) all exist on `develop` HEAD (`0be4e83`) and are **not touched by this PR** (this PR only modifies 11 files under `eventmesh-runtime/src`). The license check scans the whole tree, so this failure is pre-existing on `develop` and would affect any PR against it. The functional checks `Continuous Integration` and `Code Scanning` both pass.

qqeasonchen:
Update: `License Check` is now ✓ green. Per maintainer request, committed `97b5d41 fix(license): add Apache 2.0 license headers to 23 files flagged by CI` which adds Apache 2.0 headers (preserving shebangs / `<?xml?>` declarations) to the 23 files the workflow previously reported. All three PR-relevant checks now pass: `Continuous Integration` ✓, `Code Scanning` ✓, `License Check` ✓.

PR #5308, opened 2026-08-20
fix(cluster): unify delivery topology with Meta CAS + fencing (#5293 #5288)

### LINKED ISSUE #5288: [Bug] Start and manage membership heartbeats in sticky cluster mode

### LINKED ISSUE #5293: [Bug] Unify cluster delivery topology and fence stale partition owners
## Problem

Cluster-related code currently retains three partially overlapping models:

1. sticky local pull/local delivery configured by `EventMeshApplication.enableCluster()`;
2. partition-owned pull through `PartitionOwnership`;
3. cross-instance forwarding through `ClusterCoordinator` and `HttpForwarder`.

Their startup wiring, offset ownership, ACK semantics, failure behavior, and production maturity are not represented by one explicit topology choice. `PartitionOwnership` also relies on soft generation fencing and read/write Meta updates that are not a compare-and-set lease protocol.

Relevant code:

- `eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/boot/EventMeshApplication.java`
- `eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/ingress/UniIngressService.java`
- `eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/cluster/PartitionOwnership.java`
- `eventmesh-runtime/src/main/java/org/apache/eventmesh/runtime/cluster/ClusterMembership.java`

## Reproduction

1. Configure two instances against a shared MetaStore and Kafka topic with multiple partitions.
2. Introduce a network partition or MetaStore outage for one instance.
3. Continue publishing and consuming during the split.
4. Restore connectivity and compare per-partition delivery sets, duplicates, offsets, and ownership records.
5. Repeat for sticky mode and partition-owned mode.

## Expected behavior

The deployment must select one explicit delivery topology with documented guarantees. Ownership handoff must prevent stale owners from continuing to create valid deliveries after their epoch/lease is lost.

## Proposed direction

Abstract the cluster mode into a single configurable `DeliveryTopology` mapped from the existing `DistributionMode` enum. There is **no cross-instance forwarding**: broadcast and multicast are realized by every instance pulling the same message from the MQ and delivering it to its locally matched subscribers. The legacy `ClusterCoordinator` / `HttpForwarder` forwarding model is deprecated, not retained as a topology.

| `DistributionMode` | Delivery topology | Semantics |
| --- | --- | --- |
| `BROADCAST` / `MULTICAST` | `STICKY_LOCAL_PULL` | every instance pulls the same message from the MQ, then delivers to locally matched subscribers |
| `LOAD_BALANCE` | `PARTITION_OWNED_PULL` | each queue / partition has exactly one Runtime owner |

### `STICKY_LOCAL_PULL` (current default, from `BROADCAST` / `MULTICAST`)

- Broadcast / multicast are realized as every instance pulling the same message from the MQ; each instance delivers it to its locally matched subscribers.
- No cross-instance forwarding.
- Supports global broadcast and filter-matched delivery.
- Trade-off: MQ pulls are amplified, because every Runtime pulls the full topic.

### `PARTITION_OWNED_PULL` (high-throughput, from `LOAD_BALANCE`)

- Each queue / partition has exactly one Runtime owner.
- Requires owner fencing and handoff.
- No duplicate pulls and no cross-instance forwarding.
- Senders may optionally pass a `partitionKey` so that the same key is routed to a stable partition, preserving per-key ordering.

### `DistributionMode` cleanup

- Remove `LOAD_BALANCE_STICKY` from `DistributionMode.java`. It is no longer needed: the per-key ordering it reserved is now covered by `PARTITION_OWNED_PULL`'s optional `partitionKey`.

Each topology must document: duplicate-delivery semantics, cost / throughput characteristics, failure-recovery strategy, and a corresponding E2E test.

- Define subscription persistence, physical cursor ownership, ACK target, duplicate bound, and Meta outage behavior for each topology.
- For partition ownership, use lease revision/CAS plus owner epoch fencing; carry the epoch into polling/delivery validation.
- Keep unfinished topologies experimental until their failover contract is tested.

## Acceptance criteria

- [ ] Runtime startup selects one topology explicitly; inactive topology components are not silently half-wired.
- [ ] `DistributionMode` maps unambiguously to `STICKY_LOCAL_PULL` or `PARTITION_OWNED_PULL`; neither topology uses cross-instance forwarding.
- [ ] `LOAD_BALANCE_STICKY` is removed from `DistributionMode.java`.
- [ ] Documentation defines ACK, offset, duplicate, and recovery semantics per topology.
- [ ] Partition handoff uses an atomic lease/revision or equivalent fencing mechanism.
- [ ] A stale owner cannot continue valid delivery after losing ownership.
- [ ] Fault-injection tests cover Meta outage, network split, owner crash, graceful handoff, and recovery.
- [ ] Kafka multi-instance tests verify no unintended simultaneous partition ownership.

## Suggested labels

`bug`, `need research`, `architecture`, `cluster`, `priority:p1`

---

Supersedes #5300 (Configurable DeliveryTopology for cluster delivery), which is folded into this issue's two-topology definition.

qqeasonchen:
### Motivation

Fixes #5293
Fixes #5288

The uni-architecture cluster delivery currently mixes two modes: sticky delivery with cross-instance HTTP forwarding (`HttpForwarder` + `/internal/forward` endpoints) and partition assignment without fencing. The forwarding path adds an extra network hop, hidden failure modes, and makes delivery depend on instance-to-instance reachability. At the same time, partition assignment uses read-then-write generation overwrite, which is racy: a stale instance can silently overwrite the assignment of the rightful owner after a restart or a long GC pause. Finally, `ClusterMembership.heartbeat()` is never scheduled (#5288), so partition leases expire for live instances.

### Modifications

**1. Sticky-only delivery topology (issue #5293)**

- Delete `HttpForwarder` and the `/internal/forward` + `/internal/reply-forward` endpoints in `UniHttpServer`; `reply()` on a non-owner instance now 404s. The SDK pins to one instance via `instanceUrl`, so forwarding is no longer needed.
- Remove `DistributionMode.LOAD_BALANCE_STICKY`; `LOAD_BALANCE` absorbs the sticky semantics (partition-key hash routes to one subscriber).

**2. Atomic Meta CAS fencing (issue #5293)**

- `MetaStore.tryAcquire(key, expectedOldValue, newValue)`: atomic compare-and-swap. InMemory: `ConcurrentHashMap` putIfAbsent/replace; Nacos 2.x: `publishConfigCas` with `casMd5 = MD5(expected content)`.
- `FencingToken "<bootEpoch>:<counter>"` per JVM — monotonic and restart-safe. Assignment record `/em/assignments/<topic#partition>` = `"<token>|<ownerInstanceId>"`.
- `acquireOrFence` three-branch protocol:
  - Case 1: unclaimed (or released tombstone `""`) → CAS claim
  - Case 2: still ours → sync token
  - Case 3: another owner → CAS takeover only when the owner is TTL-evicted or our token is strictly higher; otherwise we are fenced
- `releaseStale`: partitions that left our assigner share are CAS'd to a `""` tombstone so the new rightful owner can claim them on its next cycle (prevents stranding after membership churn). The tombstone is CAS-equivalent to an absent key on both MetaStore backends.

**3. Heartbeat scheduling fix (issue #5288)**

- `ClusterMembership.heartbeat()` was never scheduled; `enableCluster` now runs it every 5s and releases it on shutdown together with the partition lease.

### Tests

- `FencingTokenTest`, `InMemoryMetaStoreTest` (incl. concurrent CAS), `PartitionFencingTest` (first claim / race / restart fencing / stale CAS).
- `ClusterDeliveryFaultTest`: in-process 3-4 instance fault injection (steady-state split, crash takeover, scale-out churn, Meta partition split-brain guard, healed partition reclaim) driven by a mutable clock — no sleeps, fully deterministic.
- Removed `ClusterForwardIntegrationTest` / `NacosClusterForwardIntegrationTest` (they covered the deleted forwarding path).
- Full runtime module suite green (189 tests, 0 failures); checkstyle clean.

### Documentation

- Does this pull request introduce a new feature? (yes)
- If yes, how is the feature documented? (`docs/eventmesh-uni-architecture-redesign.md` §13.2.10 updated with the acquireOrFence protocol, release path, and fault-injection test coverage)


qqeasonchen:
**Upstream CI all green on this PR** (5 PR-triggered workflows, `9921a4f`):

| Workflow | Result | Duration |
|---|---|---|
| Continuous Integration | ✓ | (ubuntu + macOS builds) |
| Code Scanning (java + go) | ✓ | 7m19s + 1m6s |
| License Check | ✓ | (no new license issues) |
| dependency-review | ✓ | 2m34s |
| Greetings | ✓ | 25s |

`Dependabot Auto-approve` shows `startup_failure` — this is a pre-existing workflow environment issue, not caused by this PR (it fails the same way for any PR to develop).

Same config (License Check + CI + Code Scanning) was previously green on #5307 (which was merged into master as commit `b43df620d`).

